PR #29, opened 2021-07-23
reconcile gene ids after merging raw matrices with diff feature counts

jychien:


jahilton (approved):

jychien:
I created a local branch from this PR, and was able to test and clean up the code to be consistent in using 'gene_id' as Ensembl ID column in obs. In schema 1.1.0, the column was changed to 'gene_identifiers', causing set_ensembl() for cxg schema 2.0.0 to miss the column.  
However, I ran into another problem.  I ran the flattener for the merged Slide-seq dataset, and ended up with a 'Killed: 9' because of add_nan(). In order to make the raw and final matrix the same shape, there had to be lots of nan's added, and my laptop could not handle this. So, either I run it in the future on an EC2 or we see what else we see what other options there are.

jahilton:
I think you should play around with an instance type that gives enough memory for such activities - https://aws.amazon.com/ec2/instance-types/

jychien:
Left add_nan() commented out, as cxg is still deciding what to do for normalized and raw matrices with different number of features.

PR #33, opened 2021-09-27
Tools 59 update flattener for updated cxg 2.0.0

jychien:


jychien:
@jahilton reconcile_genes() and add_zero() are the more complicated functions that I've added to the script. There are a lot of changes, and also includes https://lattice.atlassian.net/browse/TOOLS-41, as I had done some development before having to make a branch for the Hammoud datasets (cxg schema 1.1.0 version). Thanks!

jahilton (approved):

PR #52, opened 2022-10-14
Tools 97 part2 update schema

jychien:
@jahilton @liaanym 
I've ran 3 of the jun-li datasets with the new cell ID logic, and so far they've passed the validator. 
I'm thinking it will be good to merge this to main in case we need to rerun the flattener next week for some reason.


scripts/flattener.py
```diff
@@ -712,6 +696,8 @@ def set_ensembl(cxg_adata, cxg_adata_raw, redundant, feature_keys):
 	rawvar_ercc = cxg_adata_raw.var.index[cxg_adata_raw.var.index.isin(ercc_df['feature_id'])]
 	cxg_adata.var.loc[var_ercc, 'feature_biotype'] = 'spike-in'
 	cxg_adata_raw.var.loc[rawvar_ercc, 'feature_biotype'] = 'spike-in'
+	del cxg_adata_raw.var['feature_biotype'] # remove feature_biotype from var and raw.var
+	del cxg_adata.var['feature_biotype']
```
jahilton, line 700: I'm surprised to see feature_biotype being set in the first place. Is this necessary or just represents a quick fix?

liaanym: When feature_biotype was not set we had issues of having excessive columns in var. Deleting it at the end fixed the issue. It was also a fix for a bug when running it on the dataset LATDF257RBX. Originally, I commented out cxg_adata.var.insert(0,  'feature_biotype', 'gene'), however the output when running the flattener on LATDF257RBX was this:

WARNING: Validation of raw layer was not performed due to current errors, try again after fixing current errors.
ERROR: Add labels error: Column 'feature_biotype' is a reserved column name of 'var'. Remove it from h5ad and try again.
Validation complete in 0:00:11.171429 with status is_valid=False

So I chose to re-comment the line and delete feature_biotype at the end of the method which fixed the error.

jahilton: Got it! Thanks for the explanation. I'm in support of the quick fix knowing that this will all get a closer look-over very soon

jychien:
Thanks for the review! @jahilton  :) PR has been merged.

PR #60, opened 2023-01-19
removed Seurat conversion and any rpy2 dependencies

jychien:
@jahilton Here is the PR


scripts/flattener.py
```diff
@@ -870,18 +818,7 @@ def main(mfinal_id):
 	file_ext = file_url.split('.')[-1]
 	mfinal_local_path = '{}/{}.{}'.format(tmp_dir, mfinal_obj['accession'], file_ext)
 	mfinal_adata = None
```
jahilton, line 820: Can remove this line, right?

jychien: Yes! Let's remove that.

jychien: Removed in latest commit

jahilton (approved):

PR #62, opened 2023-01-31
Tools 108 filter atac ids and update to concat

jychien:
@jahilton: Things updated:
- replace append() with concat()
- make sure var index is same size as one of raw matrices
- convert author_cell_type_column datatype if needed
- atac gene activity matrices will filter ensembl IDs for appropriate annotation version

I actually don't make sure that the atac processed matrix file is indexed by ensembl. Should I make this assumption? Or do a check?

### DIFF
--- a/scripts/flattener.py
+++ b/scripts/flattener.py
@@ -957,6 +957,8 @@ def main(mfinal_id):
 			collapse_dict = reconcile_results[3]
 		else:
 			cxg_adata_raw = cxg_adata_lst[0].concatenate(cxg_adata_lst[1:], index_unique=None, join='outer')
+			if cxg_adata_raw.var.shape[0] != cxg_adata_lst[0].var.shape[0]:
+				sys.exit('There should be the same genes for raw matrices if only a single genome annotation')
 		cxg_adata_raw = cxg_adata_raw[mfinal_cell_identifiers]
 		if cxg_adata_raw.shape[0] != mfinal_adata.shape[0]:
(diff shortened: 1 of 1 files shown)


jahilton:
>I actually don't make sure that the atac processed matrix file is indexed by ensembl. Should I make this assumption? Or do a check?

I think you can assume that. It's something that we can build into checkfiles now that we have enough consistency to run these files through.

scripts/flattener.py
```diff
@@ -629,22 +633,15 @@ def set_ensembl(cxg_adata, cxg_adata_raw, redundant, feature_keys):
 		unique_to_norm =  set(cxg_adata.var.index.to_list()).difference(set(cxg_adata_raw.var.index.to_list()))
 		if len(unique_to_norm) > 0:
 			print("WARNING: normalized matrix contains Ensembl IDs not in raw: {}".format(unique_to_norm))
+	del cxg_adata_raw.var['feature_biotype'] # remove feature_biotype from var and raw.var
+	del cxg_adata.var['feature_biotype']
```
jahilton, line 637: can we prevent these (`feature_biotype`) from being added in the first place?

jychien: Yes, I want to clean up feature_biotype, but I was thinking to do it in a separate ticket to break up the work. I can make another ticket for that and also move more code from main() to separate functions.

jahilton: works for me

jahilton (approved):

PR #64, opened 2023-02-07
Tools 109 clean up var

jychien:
@jahilton 

- Retain more information from contributor anndata 
- cleaned up feature_biotype
- set global variables

### DIFF
--- a/scripts/flattener.py
+++ b/scripts/flattener.py
@@ -429,7 +429,7 @@ def compile_annotations(files):
 			sys.exit('Failed to find file {} on s3'.format(file_obj.get('@id')))
 		else:
 			print("Downloading reference: {}".format(files[key]))
-		df = pd.read_csv(filename, names=['feature_id','symbol','num'])
+		df = pd.read_csv(filename, names=['feature_id','symbol','num'], dtype='str')
 		ids  = pd.concat([ids,df])
 	return ids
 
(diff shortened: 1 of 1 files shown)


scripts/flattener.py
```diff
@@ -583,26 +591,21 @@ def add_zero(cxg_adata, cxg_adata_raw):
 		new_var.loc[genes_add, 'feature_is_filtered'] = True
 		new_adata = ad.AnnData(X=new_matrix, obs=cxg_adata.obs, var=new_var, uns=cxg_adata.uns, obsm=cxg_adata.obsm)
 		new_adata = new_adata[:,cxg_adata_raw.var.index.to_list()]
-		return(new_adata)
+		new_adata.var = new_adata.var.merge(cxg_adata.var, left_index=True, right_index=True, how='left')
+		cxg_adata = new_adata
 	else:
 		cxg_adata.var['feature_is_filtered'] = False
-		return(cxg_adata)
+		cxg_adata = cxg_adata
```
jahilton, line 598: does this do anything?

jychien: Ah, thank you for catching that. I believe I was thinking that if I had changed cxg_adata.var that I would need to reset the global variable again. But then I figured out the issue was that I needed to declare the global variables in main(), and I forgot to clean up this part. I just double checked with LATDF190KNY, and passed validator.

PR #65, opened 2023-02-15
Tools 6 geo submission and s3 uris

jychien:
@jahilton Here is the geo metadata script. I haven't thoroughly tested on other datasets in the Lattice, but it works for LATDS977NKW. Let me know if additional testing is needed

### DIFF
--- a/scripts/geo_metadata.py
+++ b/scripts/geo_metadata.py
@@ -228,11 +228,17 @@ def get_dataset(dataset):
 	desc = results.get('description','')
 
 	all_tissues = []
+	all_assays = []
 	for lib in results.get('libraries'):
 		for tissue in lib.get('biosample_summary'):
 			if tissue not in all_tissues:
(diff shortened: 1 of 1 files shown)


scripts/geo_metadata.py
```diff
@@ -0,0 +1,552 @@
...
+import pandas as pd
+import shutil
+import sys
+import scanpy as sc
+import re
+import subprocess
+from urllib.parse import urljoin
+import requests
+import gspread
+import numpy as np
```
jahilton, line 14: I think these aren't needed
anndata
os
shutil
scanpy
re
subprocess
gspread

jychien: Thanks! That is definitely a lot of unused libraries. I've removed them and also removed a line of code that converted type that wasn't necessary.

jahilton (approved):

PR #68, opened 2023-03-08
refactor gene reconciliation, remove redundant genes from all dataset…

jychien:
… and no longer collapse for dataset with >1 annot version

@jahilton 
Here is the PR. Let me know if there is anything amiss.

### DIFF
--- a/scripts/flattener.py
+++ b/scripts/flattener.py
@@ -124,6 +124,7 @@ mfinal_obj = None
 mfinal_adata = None
 cxg_adata = None
 cxg_adata_raw = None
+cxg_obs = None
 
 EPILOG = '''
 Examples:
@@ -597,7 +598,7 @@ def add_zero():
 		cxg_adata.var['feature_is_filtered'] = False
 
 
-# Use cxg_adata_raw var to map ensembl IDs and use that as index, filter against ref_files[]
+# Use cxg_adata_raw var to map ensembl IDs and use that as index and drop redundants
 # Make sure the indices are the same order for both anndata objects & clean up var metadata
 # WILL NEED TO ADD NEW BIOTYPE FOR CITE-SEQ
 def set_ensembl(redundant, feature_keys):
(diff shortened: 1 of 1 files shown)


scripts/flattener.py
```diff
@@ -607,24 +608,22 @@ def set_ensembl(redundant, feature_keys):
...
+			logging.info('drop_unmapped\t{}'.format(drop_unmapped))
+
 			cxg_adata.var = pd.merge(cxg_adata.var, cxg_adata_raw.var, left_index=True, right_index=True, how='left', copy = True)
 			cxg_adata.var = cxg_adata.var.set_index('gene_ids', drop=True)
 			cxg_adata_raw.var  = cxg_adata_raw.var.set_index('gene_ids', drop=True)
 			cxg_adata.var.index.name = None
 			cxg_adata_raw.var.index.name = None
+
+			# Drop redundant by Ensembl ID
+			cxg_adata = cxg_adata[:, [i for i in cxg_adata.var.index.to_list() if i not in redundant]]
```
jahilton, line 625: Should we log this similar to drop_unmapped?

scripts/flattener.py
```diff
@@ -656,24 +655,18 @@ def clean_var():
...
+	mfinal_adata_genes = mfinal_adata.var.index.to_list()
 	redundant = []
-	redundant_within_version = []
-	chain_redundant_genes = []
-	gene_multi_ensembl = []
-	versions_checked = []
-	genes_to_collapse_final = {}
-	collapsed_need_to_only_switch_id = {}
-	total_genes = []
+	multiple_ensembl = []
```
jahilton, line 664: I think that this could get cleaned up a little
`stats = {}`

then later, instead of the 3x `extend`, you could define the key,value
`stats['multiple_ensembl'] = multiple_ensembl`

then, you just need to adjust this line
`all_remove = list(set(redundant + multiple_ensembl + multiple_symbols))`
and call the dict
`all_remove = list(set(stats['redundant'] + stats['multiple_ensembl'] + stats['multiple_symbols']))`
or even use list comprehension
`all_remove = list(set([i for l in stats.values() for i in l]))`

scripts/flattener.py
```diff
@@ -688,85 +681,40 @@ def reconcile_genes(mfinal_obj, cxg_adata_lst, mfinal_adata_genes):
...
-	stats['redundant'] = list(set(redundant))
 	for key in stats:
 		stats[key] = set(stats[key])
 		overlap_norm = set(mfinal_adata_genes).intersection(stats[key])
 		logging.info("{}\t{}\t{}\t{}\t{}".format(key, len(stats[key]), len(overlap_norm), overlap_norm, stats[key]))
 
-	return cxg_adata_raw_ensembl, genes_to_collapse_final, redundant, genes_to_collapse_dict
+	all_remove = list(set(redundant + multiple_ensembl + multiple_symbols))
+
+	return cxg_adata_raw_ensembl, redundant, all_remove
```
jahilton, line 717: if going with the above suggestion, would need to update this to `stats['redundant']`

scripts/flattener.py
```diff
@@ -961,16 +966,14 @@ def main(mfinal_id):
...
+		# If raw matrices are annotated to multiple gencode versions, concatenate on ensembl ID and remove ambiguous symbols
 		if len(mfinal_obj.get('genome_annotations',[])) > 1:
-			reconcile_results = reconcile_genes(mfinal_obj, cxg_adata_lst, mfinal_adata.var.index.to_list())
+			reconcile_results = reconcile_genes(cxg_adata_lst)
 			cxg_adata_raw = reconcile_results[0]
-			collapse = reconcile_results[1]
-			redundant = reconcile_results[2]
-			collapse_dict = reconcile_results[3]
+			redundant = reconcile_results[1]
+			all_remove = reconcile_results[2]
```
jahilton, line 975: combine these lines to `cxg_adata_raw, redundant, all_remove = reconcile_genes(cxg_adata_lst)`?

scripts/flattener.py
jychien: Good point. For some reason, I got it in my head that you needed to instantiate the value for each key if it is a list. I think I am getting confused with a nested dictionary.
And after looking at this comment, I realized I made a mistake. It doesn't change the outcome, but 'redundant' is actually a list of Ensembl IDs. So, it is dropped separately at a different step after the Ensembl IDs have been mapped. 

I'll make all the changes:
- drop 'redundant', 'multiple_ensembl', 'multiple_symbols' variables
- Just instantiate stats = {}
- all_remove should not include 'redundant'

scripts/flattener.py
jychien: Ooh, nice, much better. I'll give it a try.

scripts/flattener.py
jychien: Sounds good.

jychien:
@jahilton I've push another commit with the changes.

jahilton (approved):

jychien:
@jahilton 
Added logging and made 'redundant' list unique

jahilton (approved):

PR #70, opened 2023-03-21
add logic for PAR_Y genes so they do not get filtered

jychien:
@jahilton

### DIFF
--- a/scripts/flattener.py
+++ b/scripts/flattener.py
@@ -718,7 +718,10 @@ def reconcile_genes(cxg_adata_lst):
 	# Log redundant gene Ensembl IDs from normalized matrix within a single version
 	for col in gene_pd_ensembl.columns:
 	    for gene in [i for i, c in collections.Counter(gene_pd_ensembl[col].dropna().to_list()).items() if c > 1]:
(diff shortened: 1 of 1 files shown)


scripts/flattener.py
```diff
@@ -718,7 +718,10 @@ def reconcile_genes(cxg_adata_lst):
 	# Log redundant gene Ensembl IDs from normalized matrix within a single version
 	for col in gene_pd_ensembl.columns:
 	    for gene in [i for i, c in collections.Counter(gene_pd_ensembl[col].dropna().to_list()).items() if c > 1]:
-	        redundant.extend(gene_pd_ensembl[gene_pd_ensembl[col] == gene].index.to_list())
+			if True in gene_pd_ensembl[gene_pd_ensembl[col]==gene].index.str.endswith("PAR_Y"):
+				redundant.extend([i for i in gene_pd_ensembl[gene_pd_ensembl[col]==gene].index.to_list() if i.endswith('PAR_Y')])
+		else:
```
jahilton, line 723: This line looks like it needs to be indexed 1 more so it lines up with the `if...`

jychien: Oh man, thank you for catching that. I can't believe I left it like that. I've cleaned up the spacing.

jahilton (approved):

PR #83, opened 2023-06-16
Adding index = False to pd.to_csv

Jchaffer787:
Trying to solve issue with 'ValueError: Length mismatch: Expected axis has 7 elements, new values have 6 elements'.


jychien:
It's running for me now. Thanks! Looks good.

PR #84, opened 2023-06-20
Tools 128 add cellculture fields

jychien:
@jahilton ' (cell culture)' or ' (organoid)' suffix is already added in get_value(), so just had to remove any additional code that looked at cell slims. Ran the flattener on: LATDF257RBX, LATDF405BWD, LATDF994BQY and tissue ontology is now endothelial cell of umbilical vein, preadipocyte, and brown preadipocyte, respectively.


jahilton (approved):
LGTM

PR #86, opened 2023-07-06
Tools 127 replace deprecated concatenate append

Jchaffer787:
Added a concat_list function which allows for concatenating a list with Anndata .concat to replace deprecated .concatenate. 
Replaced deprecated .append with .concat, changing row_to_add to a series in order to work.


jahilton (approved):

jychien:
Looks good to me!

PR #91, opened 2023-08-01
TOOLS-133-update-api-key-path

corismall:


cellxgene_resources/upload_local.ipynb
```diff
@@ -26,8 +26,8 @@
    "metadata": {},
    "outputs": [],
    "source": [
-    "scc_repo_loc = os.path.expanduser('~/GitHub/CZI/')\n",
-    "api_key_file_path = os.path.expanduser('~/Documents/cxg-api-key-dev.txt')"
+    "scc_repo_loc = os.path.expanduser('~/GitClones/Lattice-Data/')\n",
```
jahilton, line 29: I think this should be `~/GitClones/CZI` since it is the single-cell-curation repo from the CZI organization, not one of ours

cellxgene_resources/upload_local.ipynb
```diff
@@ -26,8 +26,8 @@
...
+    "api_key_file_path = os.path.expanduser('~/Documents/keys/cxg-api-key-dev.txt')"
```
jahilton, line 30: I think we'll be using production 90% of the time, so let's default to production keys/site
so `cxg-api-key.txt`

and then remove the `env='dev'` below

jahilton (approved):

PR #94, opened 2023-08-23
Tools 87 allow layers transfer of processedmatrices

Jchaffer787:


scripts/flattener.py
```diff
@@ -716,6 +718,12 @@ def add_zero():
 		new_var['feature_is_filtered'] = False
 		new_var.loc[genes_add, 'feature_is_filtered'] = True
 		new_adata = ad.AnnData(X=new_matrix, obs=cxg_adata.obs, var=new_var, uns=cxg_adata.uns, obsm=cxg_adata.obsm)
+		if cxg_adata.layers:
+			for layer in cxg_adata.layers:
+				print(cxg_adata.layers[layer].getformat())
```
jychien, line 723: Was this print statement for your QA?

scripts/flattener.py
```diff
@@ -716,6 +718,12 @@ def add_zero():
...
+				cxg_adata.layers[layer] = sparse.csr_matrix(cxg_adata.layers[layer])
```
jychien, line 723: I am thinking it should convert to sparse only if it is not sparse. And then I realized the flattener makes this check on line 1382 (original line 1375). Maybe convert adata.X and adata.layers to sparse all at that part of the code.

Jchaffer787: I have moved the check/conversion of cxg_adata.layers layers to where layers are originally checked for which is right below where adata.X is converted.

PR #95, opened 2023-09-20
Tools 138 mapped reference annotation should be from proc matrix file

Jchaffer787:


scripts/flattener.py
```diff
@@ -1223,11 +1223,13 @@ def main(mfinal_id):
...
-		# Removing mapped_reference_annotation if genome_annotations from ProcMatrixFile is empty
-		if not mfinal_obj['genome_annotations']:
-			del row_to_add['mapped_reference_annotation']
+				
 		df = pd.concat([df, row_to_add])
 		redundant = list(set(redundant))
+		
+	# Removing mapped_reference_annotation if genome_annotations from ProcMatrixFile is empty
+	if not mfinal_obj['genome_annotations']:
+		del row_to_add['mapped_reference_annotation']
```
jychien, line 1232: Thanks, Jim! The code is in the correct place, but the column `mapped_reference_annotation` actually needs to be dropped from the Dataframe called `df` . What the loop is doing in creating a Dataframe `row_to_add` (which was previously a Series prior to some code changes for concat(), and hence it's name), and concatenating it to `df`. `df` is subsequently merged into `cxg_obs` in line 1329. So, by moving it out of the mxr looping, you'll just have to drop the column once from `df`.

PR #97, opened 2023-09-26
Tools-136-curation-notebook-updates

corismall:


cellxgene_resources/curation_qa.ipynb
```diff
@@ -400,46 +400,37 @@
    "metadata": {},
    "outputs": [],
    "source": [
-    "def TENx_barcode_checker(df, num_to_check):\n",
-    "    if df.shape[0] < num_to_check:\n",
-    "        num_to_check = df.shape[0]\n",
+    "def TENx_barcode_checker(csv, num_to_check=None, barcode_to_check=None):\n",
```
jahilton, line 403: should remove the barcode_to_check parameter and dependent code - I gave an ambiguous use case

cellxgene_resources/curation_qa.ipynb
```diff
@@ -455,20 +446,10 @@
...
-    "    num_to_check = obs_df.shape[0] // 5 #default is check 20% of barcodes\n",
-    "    r = TENx_barcode_checker(obs_df, num_to_check)\n",
-    "    if r:\n",
-    "        r[prop] = a\n",
-    "        results.append(r)\n",
-    "    else:\n",
-    "        print('no barcodes checked')\n",
-    "pd.DataFrame(results).set_index(prop)"
+    "num_to_check = adata.obs.shape[0] // 5 #default is check 20% of barcodes\n",
+    "csv = '10X_barcode_table.csv'\n",
```
jahilton, line 450: we don't have a use case that would require any other csv be used, so I would remove this as a variable/function parameter and just put it directly into the function

cellxgene_resources/curation_qa.ipynb
```diff
@@ -455,20 +446,10 @@
    "metadata": {},
    "outputs": [],
    "source": [
-    "prop = 'sample_id'\n",
```
jahilton, line 458: We want to keep this general feature of checking barcodes per obs_field because it's the sample_id, donor_id, etc. that will likely be one assay or another

cellxgene_resources/curation_qa.ipynb
jahilton: Should probably also go ahead and generate the csv & include that in the PR.
The code to generate it will be convenient in the sample_code ntbk, but we can still prevent everyone from having to run it.

cellxgene_resources/curation_sample_code.ipynb
```diff
@@ -514,6 +515,144 @@
...
+    "#create barcode results dictionary\n",
+    "barcode_results = {\n",
+    "    'barcode':[],\n",
+    "    '3pv2_5pv1_5pv2':[],\n",
+    "    '3pv3':[],\n",
+    "    'multiome':[]\n",
+    "}\n",
+    "\n",
+    "#extract matrix barcodes\n",
+    "barcodes = adata.obs.index.to_series()  # this code assumes the obs index is of the n_obs barcodes\n",
```
jahilton, line 594: I wouldn't expect this task to consume any `adata` because it should just be creating a table from the v2/v3/multiome files

jahilton (changes requested):
Only looked at the barcode checker thus far

cellxgene_resources/curation_sample_code.ipynb
```diff
@@ -514,6 +515,144 @@
...
+    "<br>Note: This will need to be adapted depending on what the prefix/suffix looks like</br>"
+   ]
+  },
+  {
+   "cell_type": "code",
+   "execution_count": null,
+   "metadata": {},
+   "outputs": [],
+   "source": [
+    "# for suffix\n",
```
jahilton, line 532: I think it will be more convenient to collapse prefix & suffix into a single cell. It looks like the only distinction is the `pat`.
So you could do something like this so the user can uncomment whichever pattern is appropriate, and a little easier to see what is going on...

pattern = r'^[AGCT]*' #for suffix
#pattern = r'[ACGT]*-1$' #for prefix

adata_index_split = adata.obs.index.to_series().str.split(pat = pattern, regex=True, expand=True)
adata_index_split.iloc[:,1].value_counts(dropna=False)

cellxgene_resources/curation_sample_code.ipynb
```diff
@@ -651,6 +790,58 @@
...
+    "# Use pd.read_table to parse the gtf & apply various manipulations to produce symbol-ENSG table\n",
+    "gtf = pd.read_table('path/to/gtf', header=None)\n",
+    "gene_dict = {\n",
+    "        'gene_ids': [],\n",
+    "        'gene_symbols': []\n",
+    "    }\n",
+    "\n",
+    "# extract gene_ids and gene_symbols from last column in the gtf table\n",
+    "for index,row in gtf.iterrows():  \n",
+    "    ensg_id = row[8].split(';')[0].split('\"')[1].split('.')[0]\n",
```
jahilton, line 815: Got an error on this line on any sample GENCODE file I tried
`KeyError: 8`

jahilton: My fault - user error.

cellxgene_resources/curation_sample_code.ipynb
```diff
@@ -651,6 +790,58 @@
...
+   ]
+  },
+  {
+   "cell_type": "code",
+   "execution_count": null,
+   "metadata": {},
+   "outputs": [],
+   "source": [
+    "# Use pd.read_table to parse the gtf & apply various manipulations to produce symbol-ENSG table\n",
+    "gtf = pd.read_table('path/to/gtf', header=None)\n",
```
jahilton, line 807: I think it will be more user-friendly to define the 'path/to/gtf' separated from the rest of the code in the line. So...
gtf_file = 'path/to/gtf'
gtf = pd.read_table(gtf_file, header=None)

but trying on a gencode file, I get
`ParserError: Error tokenizing data. C error: Expected 1 fields in line 6, saw 9`
from this read_table

I think `comment='#'` parameter is needed

cellxgene_resources/curation_sample_code.ipynb
```diff
@@ -651,6 +790,58 @@
...
+    "gene_dict = {\n",
+    "        'gene_ids': [],\n",
+    "        'gene_symbols': []\n",
+    "    }\n",
+    "\n",
+    "# extract gene_ids and gene_symbols from last column in the gtf table\n",
+    "for index,row in gtf.iterrows():  \n",
```
jahilton, line 814: Will be a lot more efficient to manipulate the dataframe in place rather than export that to a dictionary.
And the print statements here are excessive. There are far too many for them to be reviewable/actionable.

cellxgene_resources/curation_sample_code.ipynb
```diff
@@ -651,6 +790,58 @@
...
+    "annotations.set_index('gene_ids', inplace=True, )\n",
+    "\n",
+    "# create dataframe from dictionary\n",
+    "annotations = pd.DataFrame(gene_dict)  \n",
+    "print(annotations.shape)\n",
+    "print(annotations['gene_symbols'].value_counts(dropna=False))\n",
+    "\n",
+    "# write dataframe to tsv file in ref_file folder\n",
+    "path = 'path/to/lattice_curation/ref_files/'\n",
+    "file = 'refdata-[source]-[ref_version]_genes_gtf.tsv'  # change reference version and source as needed\n",
```
jahilton, line 840: The refdata-* format was simply derived from an input gtf that started refdata-*
So in order to prevent the user from needing to update another variable and to maintain provenance from the gtf that this table was created from, I would just manipulate the name of the input gtf to get what you want...
`file = gtf_file.replace('.','_') + '.tsv'`
so that *.gtf just becomes *_gtf.tsv

cellxgene_resources/curation_qa.ipynb
corismall: i tried to push a zipped version of the csv, but it was too large: 
`remote:error: GH001: Large files detected. You may want to try Git Large File Storage - https://git-lfs.github.com.`

corismall: nevermind, the gzip version worked, the non-zipped file was the issue

cellxgene_resources/curation_qa.ipynb
```diff
@@ -455,19 +450,22 @@
...
-    "    print(a)\n",
     "    obs_df = adata.obs[adata.obs[prop] == a]\n",
     "    num_to_check = obs_df.shape[0] // 5 #default is check 20% of barcodes\n",
     "    r = TENx_barcode_checker(obs_df, num_to_check)\n",
-    "    if r:\n",
-    "        r[prop] = a\n",
-    "        results.append(r)\n",
-    "    else:\n",
-    "        print('no barcodes checked')\n",
+    "    print(r)\n",
```
jahilton, line 460: Should remove this pre statement as it muddies up the overall report

cellxgene_resources/curation_qa.ipynb
```diff
@@ -400,46 +400,41 @@
...
+   ]
+  },
+  {
+   "cell_type": "code",
+   "execution_count": null,
+   "metadata": {},
+   "outputs": [],
+   "source": [
+    "def TENx_barcode_checker(obs_df, num_to_check):\n",
+    "    ref_df = pd.read_csv('../cellxgene_resources/10X_barcode_table.csv', sep=',', header=0, index_col='barcode')\n",
```
jahilton, line 424: the csv can go into cellxgene_resources/ref_files
and since the qa notebook can be assumed to be run from cellxgene_resources, it can be `ref_files/10X_barcode_table.csv`

cellxgene_resources/curation_qa.ipynb
```diff
@@ -400,46 +400,41 @@
...
+    "    obs_df_split = obs_df.index.str.split('([ACTG]{16})')  # starting at obs_df, we want to compare barcodes to ref_df\n",
+    "    barcodes = pd.DataFrame([b for l in obs_df_split for b in l if re.match(r\".*[ACTG]{16}.*\", b)])  # get barcodes from obs_df taking into account various barcode labels\n",
+    "    barcode_fraction = barcodes.sample(num_to_check, axis=0) \n",
+    "    barcode_fraction.rename(columns={0:'barcode'},inplace=True)\n",
+    "    barcode_fraction.set_index('barcode', inplace=True)\n",
+    "    barcode_fraction['adata_barcode'] = 1\n",
```
jahilton, line 430: Why is this needed?

cellxgene_resources/curation_qa.ipynb
```diff
@@ -400,46 +400,41 @@
...
+    "    barcode_fraction['summary'] = 'placeholder'\n",
```
jahilton, line 431: Why is this needed?

cellxgene_resources/curation_qa.ipynb
```diff
@@ -455,19 +450,22 @@
...
     "    r = TENx_barcode_checker(obs_df, num_to_check)\n",
-    "    if r:\n",
-    "        r[prop] = a\n",
-    "        results.append(r)\n",
-    "    else:\n",
-    "        print('no barcodes checked')\n",
+    "    print(r)\n",
+    "    r_dict = r['summary'].value_counts().to_dict()\n",
+    "    r_dict[prop] = a\n",
+    "    r_list = ['3pv2_5pv1_5pv2', '3pv3', 'multiome', 'multiple', 'none']\n",
```
jahilton, line 463: I think this and the following for loop are unnecessary.
You can just append `.fillna(0)` to the last line so you don't have NaNs clogging up the report

cellxgene_resources/curation_qa.ipynb
jahilton: Actually, might even want to move defining ref_df to outside the function, so the notebook loads it only once, instead of every time the function is called. (it's big so it's not a super quick load and the time could add up for big datasets)
So unde `results = []` below probably makes sense

jahilton (changes requested):

cellxgene_resources/curation_sample_code.ipynb
corismall: i believe i've fixed this: used more regex patterns/contain function -- see extract_gene_id_and symb()

cellxgene_resources/curation_qa.ipynb
corismall: it's not, was removed

cellxgene_resources/curation_qa.ipynb
corismall: removed

cellxgene_resources/curation_qa.ipynb
corismall: The last line as in: `pd.DataFrame(results).set_index(prop).fillna(0)`? Not sure how you want results formatted. Just output of TENx_barcode_checker()?

jahilton: Correct. That last line.
Full thought process...
The `r_list = `... and following loop lines take 4 lines of code to fill in 0s where you don't have any data.
If you remove those, then the last line with the DataFrame will fill in those gaps with NaNs.
Probably easier on the eyes to have those NaNs replaced with 0s, but you can take care of that with just appending `.fillna(0)` as you've noted.
So same read-out, but ditch 4 lines of code.

jahilton: and add `astype(int)` to the final line, or else you'll get everything on the table as `.0`
So last line is `pd.DataFrame(results).set_index(prop).fillna(0).astype(int)`

cellxgene_resources/curation_qa.ipynb
```diff
@@ -400,46 +400,16 @@
...
-    "                elif barcode in v3_list:\n",
-    "                    barcodes['3pv3'] += 1\n",
-    "                else:\n",
-    "                    barcodes['none'] += 1\n",
-    "        return barcodes"
+    "def TENx_barcode_checker(ref_df, obs_df, num_to_check):\n",
+    "    obs_df_sample = obs_df.sample(num_to_check, axis=0)  # can add random_state=1 for reproducibility \n",
+    "    obs_df_split = obs_df_sample.index.str.split('([ACTG]{16})')\n",
+    "    barcodes = pd.DataFrame([b for l in obs_df_split for b in l if re.match(r\".*[ACTG]{16}.*\", b)])\n",
+    "    barcodes.rename(columns={0:'barcode'},inplace=True)\n",
```
jahilton, line 407: To handle subsets of data that don't have any barcode to check, you'll need to add a check
after defining `barcodes`, add these 2 lines
```
if barcodes.empty:
    return pd.DataFrame({'summary':['no barcode'] * num_to_check})
```
Should add a row to the report table like the others, but will just report how many observations were checked for barcodes without finding one

cellxgene_resources/curation_sample_code.ipynb
```diff
@@ -514,6 +515,111 @@
...
+    "<br>Note: This will need to be adapted depending on what the prefix/suffix looks like</br>"
+   ]
+  },
+  {
+   "cell_type": "code",
+   "execution_count": null,
+   "metadata": {},
+   "outputs": [],
+   "source": [
+    "#for prefix and suffix\n",
```
jahilton, line 532: ditch this top comment

cellxgene_resources/curation_sample_code.ipynb
```diff
@@ -514,6 +515,111 @@
...
+    "barcode_table_df.to_csv('10X_barcode_table.csv.gz', sep=',', index=False, compression='gzip')"
+   ]
+  },
+  {
+   "cell_type": "code",
+   "execution_count": null,
+   "metadata": {},
+   "outputs": [],
+   "source": [
+    "# add summary column with the corresponding ref list \n",
```
jahilton, line 602: This looks like a leftover cell that should be deleted.

cellxgene_resources/curation_sample_code.ipynb
```diff
@@ -651,6 +757,92 @@
...
+   ]
+  },
+  {
+   "cell_type": "code",
+   "execution_count": null,
+   "metadata": {},
+   "outputs": [],
+   "source": [
+    "# Extract and de-duplicate gene annotations and write them to a tsv file\n",
+    "for file in gtf_list:\n",
```
jahilton, line 836: I can't think of a use case for needing to make this more high-throughput for a list of files. Typically, it's "the paper/lab says they used X.gtf so I'm going to start with that one". So unless you have a use case in mind, I would simplify by just having this execute on a single file

cellxgene_resources/curation_sample_code.ipynb
```diff
@@ -651,6 +757,92 @@
...
+    "# Filters out gene_symbols with multiple gene_ids and assigns 'multiple' as gene_id to corresponding gene_symbol\n",
+    "def assign_multiple(gtf_df):\n",
+    "    '''\n",
+    "    Input: gtf pandas dataframe with gene_id and gene_symbol columns\n",
+    "    Output: de-duplicated dataframe\n",
+    "    '''\n",
+    "    dups = gtf_df[gtf_df.duplicated(subset='gene_symbols',keep=False)]\n",
+    "    dups['gene_ids'] = 'multiple'\n",
+    "    dups = dups.drop_duplicates()  # only keep one instance of the gene_symbol\n",
+    "    gene_df_wo_dups = gene_df.drop_duplicates(subset='gene_symbols',keep=False) \n",
```
jahilton, line 813: This line throws an error for undefined `gene_df`. I think `gtf_df` is intended

cellxgene_resources/curation_sample_code.ipynb
```diff
@@ -651,6 +757,92 @@
...
+   "metadata": {},
+   "outputs": [],
+   "source": [
+    "# Extract and de-duplicate gene annotations and write them to a tsv file\n",
+    "for file in gtf_list:\n",
+    "    gtf_df = extract_gene_id_and_symb(file)\n",
+    "    gene_id_symbol = assign_multiple(gtf_df)\n",
+    "\n",
+    "    # write dataframe to tsv file in ref_file folder\n",
+    "    dir = '/path/to/cxg_curation/ref_files/'\n",
```
jahilton, line 841: I think separating out `dir` from the file name is unnecessary as I don't see any use for it to be used independently
So I would simplify things to just accept the full path in a single variable (which some notebooks are set up to autocomplete easily anyway).

cellxgene_resources/curation_sample_code.ipynb
```diff
@@ -651,6 +757,92 @@
...
+    "    tsv_file = file.replace('.','_') + '.tsv' \n",
+    "    gene_id_symbol.to_csv(dir + tsv_file, sep= '\\t', compression= 'gzip')"
```
jahilton, line 843: `'\\t'` has 1 too many `\`
Currently produces `TypeError: "delimiter" must be a 1-character string`

jahilton: Shouldn't need compression here. Not that big and it will create issues when trying to read it (that part can be tested by adding it to the list of CellRanger references to check through).

jahilton (changes requested):

cellxgene_resources/curation_qa.ipynb
corismall: I found a hiccup in the code when generating `num_to_check`. Not sure how likely this will happen, but for `num_to_check = obs_df.shape[0] // number` where the number used in floor division results in `num_to_check =0`. Where there are low numbers of barcodes associated with a particular donor/sample, but still barcodes we can check?

cellxgene_resources/curation_qa.ipynb
```diff
@@ -455,19 +425,24 @@
    "metadata": {},
    "outputs": [],
    "source": [
-    "prop = 'sample_id'\n",
+    "prop = 'donor_id'\n",
     "\n",
     "results = []\n",
+    "csv = '../ref_files/10X_barcode_table.csv'\n",
```
jahilton, line 431: This is run from cellxgene_resources/, which is where ref_files/ exists.
The `..` should be removed - it fails to find the file other wise (`..` tells it to look in one parent directory up)

cellxgene_resources/curation_qa.ipynb
jahilton: Ya, we should still check the low numbers
Although I think it's unlikely to happen often
We can always update later based on behavior of common use

corismall: i'll make a note

cellxgene_resources/curation_sample_code.ipynb
corismall: > `'\\t'` has 1 too many `\` Currently produces `TypeError: "delimiter" must be a 1-character string`

odd, i haven't come across this issue before: `sep='\t'` looks fine in both vscode and in the ipynb git preview but not raw. Not sure how to change this, will look into it.
Screenshot 2023-10-04 at 4 42 19 PM

cellxgene_resources/curation_sample_code.ipynb
```diff
@@ -651,6 +730,80 @@
...
+   "metadata": {},
+   "outputs": [],
+   "source": [
+    "# Extract and de-duplicate gene annotations and write them to a tsv file\n",
+    "gtf_df = extract_gene_id_and_symb(dir, gtf)\n",
+    "gene_id_symbol = assign_multiple(gtf_df)\n",
+    "\n",
+    "# write dataframe to tsv file in ref_file folder\n",
+    "tsv_file = gtf.replace('.','_') + '.tsv' \n",
+    "gene_id_symbol.to_csv('/path/to/cxg_curation/ref_files/' + tsv_file, sep=\"\\t\")"
```
jahilton, line 804: can replace the path to just be `ref_files/`

cellxgene_resources/curation_sample_code.ipynb
```diff
@@ -651,6 +730,80 @@
...
+    "    return gene_df_multiple"
+   ]
+  },
+  {
+   "cell_type": "code",
+   "execution_count": null,
+   "metadata": {},
+   "outputs": [],
+   "source": [
+    "# Extract and de-duplicate gene annotations and write them to a tsv file\n",
```
jahilton, line 798: probably helpful to have placeholders signaling the user to define variables
`dir = ''`
`gtf = ''`

jahilton (approved):
LGTM & works great!

PR #98, opened 2023-09-27
Tools 130 update flattener logging, fix prefix error, and clarify download printouts

Jchaffer787:
Updated flattener logging:
Errors now result in a sys.exit. Warnings do not but are printed out at the end of the script for visibility. Errors and warnings are both logged to the log_file, and the log_file is overwritten with each run of the flattener.

Fixed prefix error:
If cell_label_location is prefix, then create a list containing all the prefixes and remove the target prefix (mapping_label) from the list.
Then check the list for any prefixes that contain mapping_label, if none then perform regular .startswith() check.
If the list of prefixes does contain a prefix that contains mapping_label, then perform a more thorough .startswith() check.
This more thorough check checks if the mfinal_obj_identifier starts with the prefix, but also that it doesn’t start with any of the other prefixes.

Clarified download printouts:
The program now prints out if a file is found locally and therefore will not be downloaded. The printout that informs the user that the file is being downloaded now clarifies that the file has not been found locally.


scripts/flattener.py
```diff
@@ -843,7 +861,7 @@ def reconcile_genes(cxg_adata_lst):
 	for key in stats:
 		stats[key] = set(stats[key])
 		overlap_norm = set(mfinal_adata_genes).intersection(stats[key])
-		logging.info("{}\t{}\t{}\t{}\t{}".format(key, len(stats[key]), len(overlap_norm), overlap_norm, stats[key]))
+		warning_list.append("{}\t{}\t{}\t{}".format(key, len(stats[key]), len(overlap_norm), overlap_norm))
```
jychien, line 864: For line 850: `stats['redundant'] = redundant`, it is adding a list of Ensembl IDs that map to redundant gene symbols. And when I flatten LATDF695OQI, I am getting this in the log:

> redundant	110	0	set()
> WARNING: 8 redundant gene_ids dropped:	['ENSG00000180425', 'ENSG00000256806', 'ENSG00000196139', 'ENSG00000112096', 'ENSG00000105519', 'ENSG00000128655', 'ENSG00000196642', 'ENSG00000150076']

The first line is confusing because the for loop in line 861 is checking for all redundant ID that overlap the gene symbols in mfinal_adata_genes. This will always be 0, since mfinal_adata_genes is gene symbol and the redundant  is Ensembl ID. (Can you confirm this?). I think the first line is confusing and should be removed. So, I would either drop `stats['redundant'] = redundant`, or else the other option is to select for it in the `for key in stats:` loop, and not print the last two elements in the format parameters.

Hopefully that makes sense. Otherwise, looks great!!

Jchaffer787: > > This will always be 0, since mfinal_adata_genes is gene symbol and the redundant is Ensembl ID. (Can you confirm this?).

Can confirm.
I'll go ahead and drop `stats['redundant'] = redundant` then.

Jchaffer787:
Okay got the commit to the right branch this time 😁 . I accidentally committed to main so I just made another commit replacing the line. Hopefully thats okay.

jychien:
Not a problem! :) Looks great! Read to merge into main! If you slack out a quick summary of changes to Jennifer/Erica, that would be great. Thanks!

PR #102, opened 2023-10-17
Tools 139 suppressing unnessecary warnings

Jchaffer787:
Simple fix, not actually suppressing any warnings, but instead now printing them out exclusively in the logging outfile. This should make the print window much less crowded so that the downloads are easier to follow, while also not losing any important contextual information.


jychien:
Wow! So much better! Great find, Jim! Looks very good, let's merge.

PR #104, opened 2023-10-30
Tools 144 check primary portion field in proc matrix file

Jchaffer787:
Added checks for ProcessedMatrixFile's ['primary_portion'], specifically whether 'obs_value' is an actual obs column, and that the 'values' are all present within that column.

### DIFF
--- a/scripts/flattener.py
+++ b/scripts/flattener.py
@@ -1421,7 +1421,7 @@ def main(mfinal_id):
 		cxg_obs = pd.merge(cxg_obs, df[['disease_ontology_term_id', 'reported_diseases', 'sex_ontology_term_id']], left_on="raw_matrix_accession", right_index=True, how="left" )
 
 	# Check that primary_portion.obs_field of ProcessedMatrixFile is present in cxg_obs
-	if not mfinal_obj.get('primary_portion', None): # Checking for presence of 'primary_portion'
+	if mfinal_obj.get('primary_portion', None): # Checking for presence of 'primary_portion'
 		primary_portion = mfinal_obj.get('primary_portion')
 		if primary_portion.get('obs_field') not in cxg_obs.columns:
 			logging.error("ERROR: 'obs_field' value '{}' not found in cxg_obs columns".format(primary_portion.get('obs_field')))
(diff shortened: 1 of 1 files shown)


scripts/flattener.py
```diff
@@ -1420,6 +1420,18 @@ def main(mfinal_id):
...
 		cxg_obs = pd.merge(cxg_obs, df[['disease_ontology_term_id', 'reported_diseases', 'sex_ontology_term_id']], left_on="raw_matrix_accession", right_index=True, how="left" )
 
+	# Check that primary_portion.obs_field of ProcessedMatrixFile is present in cxg_obs
+	primary_portion = mfinal_obj.get('primary_portion')
+	if primary_portion.get('obs_field') not in cxg_obs.columns:
+		logging.error("ERROR: 'obs_field' value '{}' not found in cxg_obs columns".format(primary_portion.get('obs_field')))
+		sys.exit("ERROR: 'obs_field' value '{}' not found in cxg_obs columns".format(primary_portion.get('obs_field')))
+
+	# Check that all primary_portion.values of ProcessedMatrixFile are found in the 'obs_field' column of cxg_obs
+	missing = [f for f in primary_portion.get('values') if f not in cxg_obs[primary_portion.get('obs_field')]]
```
jychien, line 1430: Looks like there is an error. `cxg_obs[primary_portion.get('obs_field')]` returns a series, you need it to be a list.

scripts/flattener.py
```diff
@@ -1420,6 +1420,18 @@ def main(mfinal_id):
 		get_sex_ontology(df)
 		cxg_obs = pd.merge(cxg_obs, df[['disease_ontology_term_id', 'reported_diseases', 'sex_ontology_term_id']], left_on="raw_matrix_accession", right_index=True, how="left" )
 
+	# Check that primary_portion.obs_field of ProcessedMatrixFile is present in cxg_obs
+	primary_portion = mfinal_obj.get('primary_portion')
```
jychien, line 1424: `primary_portion` is not found in all objects, and will error with, for example if you flatten LATDF512YHM

jychien: Can be changed to` primary_portion = mfinal_obj.get('primary_portion', None)`. This returns None if there is no such field.

jychien:
Looks good to me! Ready to merge to main!

PR #107, opened 2023-11-08
update documentation on installing ncftp

jychien:
Update documentation on installing ncftp on EC2. Thanks for the review @Jchaffer787


Jchaffer787:
Was a great call to put the correct download link into the ticket once we found it.
Looks good to me!

jychien:
Thanks, Jim! I am thinking the future us will be thankful :P

Jchaffer787 (commented):
LGTM!

PR #109, opened 2023-11-10
update documentation for environment config

brian-mott:
macOS and linux need python-magic instead of python-magic-bin


jahilton (approved):
LGTM, and it runs as expected

PR #111, opened 2023-12-07
Tools 139 updates for schema 4.0

Jchaffer787:
Various updates in order to transition the flattener over to schema 4.0:

Adding tissue_type field to obs.

Making it so self_reported_ethnicity_ontology_term_id can be a list.

Adding a check for reserved uns names.

Adding a check that adata.X and adata.raw.X are the correct datatypes.

Adding feature_length to var.

### DIFF
--- a/scripts/flattener.py
+++ b/scripts/flattener.py
@@ -1446,6 +1446,11 @@ def main(mfinal_id):
 	var_meta = mfinal_adata.var.select_dtypes(include=keep_types)
 	cxg_adata = ad.AnnData(mfinal_adata.X, obs=cxg_obs, obsm=cxg_obsm, var=cxg_var, uns=cxg_uns)
 	cxg_adata.var = cxg_adata.var.merge(var_meta, left_index=True, right_index=True, how='left')
+	
+	# Removing feature_length column from var if present
+	if 'feature_length' in cxg_adata.var.columns:
+		adata.var.drop(columns=['feature_length'], inplace=True)
+		
 	if not sparse.issparse(cxg_adata.X):
 		cxg_adata.X = sparse.csr_matrix(cxg_adata.X)
 	elif cxg_adata.X.getformat()=='csc':
--- a/scripts/flattener.py
+++ b/scripts/flattener.py
@@ -33,7 +33,7 @@ cell_metadata = {
 		'donor_id',
 		'age_display',
 		'sex',
-		'summary_ethnicity',
+		'ethnicity',
 		'causes_of_death.term_name',
 		'diseases.term_id',
 		'diseases.term_name',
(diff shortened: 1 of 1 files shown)


jahilton (commented):
I don't see any place that tissue_type is set
Everything else looks good

scripts/flattener.py
```diff
@@ -144,7 +145,7 @@
 	'donor_sex': 'sex',
```
jychien, line 145: @Jchaffer787 There needs to be an entry added to property map. Right now the field being outputted is called `sample_@type`. So, you'll need to add `'sample_@type': 'tissue_type',`

scripts/flattener.py
```diff
@@ -341,7 +342,11 @@ def gather_objects(input_object, start_type=None):
 def get_value(obj, prop):
 	path = prop.split('.')
 	if len(path) == 1:
-		return obj.get(prop, unreported_value)
+		if path[0] == '@type':
```
jychien, line 345: We just need to remember that if we end up ever adding `@type` to any of the other object types (donor, tissue section, suspension, library, raw_matrix, or seq_run), this code will need to have a more specific if clause to distinguish.

jychien:
The value to the new tissue type field is currently `Tissue`. This needs to be mapped to lower case `tissue`. I'm not sure why github is not letting me comment in line of the code, but you'll need to add some code in `clean_obs()`. And probably good to put in a check:

	valid_tissue_types = ['tissue', 'organoid', 'cell culture']
	cxg_obs['tissue_type'] = cxg_obs['tissue_type'].str.lower()
	for i in cxg_obs['tissue_type'].unique().tolist():
		if i not in valid_tissue_types:
			logging.error('ERROR: not a valid tissue type:\t{}'.format(i))
			print('ERROR: not a valid tissue type:\t{}'.format(i))

scripts/flattener.py
```diff
@@ -433,6 +447,32 @@ def gather_pooled_metadata(obj_type, properties, values_to_add, objs):
...
+				else:
+					ethnicity_list.sort()
+					value = ','.join(ethnicity_list)
+					latkey = (obj_type + '_' + prop).replace('.','_')
+					key = prop_map.get(latkey, latkey)
+					values_df.loc[key,ident] = value
+			for i in unknowns:
+				values_df[i] = 'unknown'
+			for index, row in values_df.iterrows():
+				values_to_add[index] = 'pooled [{}]'.format(','.join(row.to_list()))
```
jychien, line 472: For LATDF477OUM, this is returning `pooled [HANCESTRO:0005,HANCESTRO:0005]` for self_reported_ethnicity_ontology_term_id. Was there a specific ProcMatrixFile that you were targeting this format for?
This is a little tricky. I think for these, where there are samples are multiplexed but not computationally de-multiplexed, if there ends up being a list of ethnicities, the actual ethnicity should be `unknown`. @jahilton Thoughts? A single cell from that multiplexed sample can be from either donor. And if the donors are of different ethnicities, there is no way to determine which ethnicity that cell is. 
So, for this code, I think first check uniqueness of list. And then, if this pooled sample is still of mixed ethnicities, then set to `unknown`.

jahilton: So we're thinking of the case where a cell is from multiple donors?
Ideally, if the ethnicities match completely amongst those donors, then we use those (whether it's one or a list)
If there is any different amongst them, then we should go to `unknown`
But if the comparison is too onerous, then just pooling --> `unknown` always would be acceptable, I think

Thanks for flagging that case

jychien:
Sorry, I forgot there is a space in `cell culture`. I am getting the following error for LATDF994BQY:

> ERROR: not a valid tissue type: cellculture

jychien:
I ran through 10 of the 11 dataset in WRN-644 (cannot flatten peak matrix), and they all passed the schema 4.0 validator. Looks good to go! @Jchaffer787 Thank you for all the work! Let's go ahead and merge to main!

PR #112, opened 2023-12-11
Tools-150 annotation map from gtf (new)

corismall:


cellxgene_resources/curation_sample_code.ipynb
```diff
@@ -747,50 +747,54 @@
...
+    "            df1 = feature_info.str.split(';', expand = True).dropna(axis = 1)\n",
+    "            print('Need to id gene_symbols and ensg_id cols, here\\'s the first row:\\n')\n",
+    "            for k,v in df1.head(1).to_dict('records')[0].items():\n",
+    "                print(str(k),':',v)  # caution: relevant columns with gene symbols and ensg ids may change for the specific gtf\n",
+    "            gene_id_col = int(input('Column number for ensg_ids?\\n'))\n",
+    "            gene_symbol_col = int(input('Column number for gene_symbols?\\n'))\n",
+    "\n",
+    "            for p in ens_id_prefixes:\n",
+    "                if p in df1[gene_id_col][0]:\n",
+    "                    print(f'Extracting {p} ids...')\n",
```
jahilton, line 777: I think you could remove this print statement.
Right now, you enter a few options and then you see this that ends with `...` so you kinda wait for something after that.

jahilton (changes requested):
Not sure how but testing 2020-A CellRanger gtf, there's a couple g
ENSG00000269466 should be H3.Y but it is H3 in the tsv produced
ENSG00000235300 should be THRA1/BTR but is is THRA1 in the tsv produced

cellxgene_resources/curation_sample_code.ipynb
```diff
@@ -748,29 +748,21 @@
    "metadata": {},
    "outputs": [],
    "source": [
-    "# Extracts gene_id and gene_symbol from gtf\n",
+    "#Read in Gencode-formatted gtf file\n",
+    "genome_version = ''\n",
+    "ref_file = f's3://submissions-lattice/reference-files/refdata-gex-{genome_version}.gtf'\n",
```
jahilton, line 753: The GENCODE-formatted gtf files won't have consistent naming conventions (only 2 of the CellRanger outputs would fit the proposed pattern). And they will likely be local, not read directly from S3.
So would suggest:

- removing `genome_version` line
- moving `my_dir = 'path/to/working/dir'` up here (`dir` is a built-in function so I'd avoid using that variable)
- then `ref_file = my_ref.gtf` (as a placeholder, like the dir)

cellxgene_resources/curation_sample_code.ipynb
```diff
@@ -748,29 +748,21 @@
...
+    "column_names = ['seqname','source','feature','start','stop','score','strand','frame','expand_me']\n",
+    "gtf_df = pd.read_table(ref_file, names=column_names, comment='#')\n",
```
jahilton, line 755: following the above suggestions, would need to edit to `my_dir + ref_file` so it can be found

cellxgene_resources/curation_sample_code.ipynb
```diff
@@ -800,16 +785,10 @@
...
-    "gtf_df = extract_gene_id_and_symb(dir, gtf)\n",
-    "gene_id_symbol = assign_multiple(gtf_df)\n",
-    "\n",
-    "# write dataframe to tsv file in ref_file folder\n",
-    "tsv_file = gtf.replace('.','_') + '.tsv' \n",
-    "gene_id_symbol.to_csv('ref_files/' + tsv_file, sep=\"\\t\")"
+    "# Write gene_df to tsv file\n",
+    "dir = 'path/to/working/dir'\n",
+    "tsv_file = ref_file.split('/')[4].replace('.','_') + '.tsv'\n",
+    "gene_df.to_csv(dir + tsv_file, sep=\"\\t\")"
```
jahilton, line 791: Following above suggestions:

- the `dir = ` line is removed as it is addressed above
- the tsv_file naming gets much simpler (`ref_file.replace('.','_') + '.tsv'`)
- update the `dir` --> `my_dir`
- in the `sep`, I'd use single quotes for consistency with the rest of the code
- add `index=False` to the `to_csv()`

jahilton (changes requested):
Just a few minor updates, I thinik

jahilton (approved):
LGTM

PR #115, opened 2024-01-22
refactor reported diseases and also allow for exp disease to be a list

jychien:
Allow for experimental_variable_disease as a list. @jahilton  
Jira ticket: https://lattice.atlassian.net/browse/TOOLS-158

### DIFF
--- a/scripts/flattener.py
+++ b/scripts/flattener.py
@@ -653,11 +653,14 @@ def quality_check(adata):
 			logging.error('ERROR: There are more genes in normalized genes than in raw matrix.')
 			sys.exit("ERROR: There are more genes in normalized genes than in raw matrix.")
 
-
 # Return value to be stored in disease field based on list of diseases from donor and sample
 def clean_list(lst, exp_disease):
 	lst = lst.split(',')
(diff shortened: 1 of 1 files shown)


scripts/flattener.py
```diff
@@ -653,11 +653,14 @@ def quality_check(adata):
 			logging.error('ERROR: There are more genes in normalized genes than in raw matrix.')
 			sys.exit("ERROR: There are more genes in normalized genes than in raw matrix.")
 
-
 # Return value to be stored in disease field based on list of diseases from donor and sample
 def clean_list(lst, exp_disease):
 	lst = lst.split(',')
-	disease = exp_disease['term_id'] if exp_disease['term_id'] in lst else 'PATO:0000461'
+	exp_disease_list = [i['term_id'] for i in exp_disease]
+	if [i for i in exp_disease_list if i in lst]:
```
jahilton, line 660: Think we should have a warning here if this is longer than 1, just so the wrangler is aware that they see 1 value, but multiple were found

jychien: Sounds good! I've updated the flattener and added a variable `disease_found` so that the code is easier to read. https://lattice.atlassian.net/browse/TOOLS-158?focusedCommentId=17679

PR #116, opened 2024-01-30
Tools 154 validate uns colors

Jchaffer787:
Adding checks for if _colors columns in uns are invalid. If invalid an error logging is displayed in the command prompt window.

Checks are as follows:
1. Check that corresponding obs column exists

2. Check if ‘_colors’ is of None type / not a numpy array

3. Check if ‘_colors’ contains non-string values

4. Check if ‘_colors’ has less entries than unique values in the associated obs column

5. Check if ‘_colors’ is not all hex / is not all CSS4 named colors

Jenny found an issue with the error logging for tissue type, and the presence of a space in ‘cellculture’ in valid_tissue_types.
The space has been removed so that valid_tissue_types now contains 'cellculture' instead of 'cell culture'.

I also added a quick fix to the warning output for an uns key being dropped so it now says what key was dropped.


jychien:
Looks great! Thanks, Jim!

PR #118, opened 2024-02-09
refactor pooled ethnicity

jychien:
@Jchaffer787  I created the wrong PR previously. Here is the PR for my commit (and not Jason's commits that I was behind on my local branch).
Here is what I've tested: https://lattice.atlassian.net/browse/TOOLS-155?focusedCommentId=17851. Thanks!


Jchaffer787:
I've re-ran the tests that were mentioned on the Jira ticket and had the same results.
Looks good to me!

PR #119, opened 2024-02-09
Adding code for logging changes

Jchaffer787:
Modified the existing warning output so that it only prints 10 of the unmapped genes, and states that the full list is available in the logging file.

Added handling of warning list so that full list of unmapped gene_ids is only printed to the logging file. Also made it so that the 'preview' list is only printed to the command prompt, and not also printed to the logging file

### DIFF
--- a/scripts/flattener.py
+++ b/scripts/flattener.py
@@ -960,7 +960,8 @@ def reconcile_genes(cxg_adata_lst):
 	for key in stats:
 		stats[key] = set(stats[key])
 		overlap_norm = set(mfinal_adata_genes).intersection(stats[key])
-		warning_list.append("{}\t{}\t{}\t{}".format(key, len(stats[key]), len(overlap_norm), overlap_norm))
+		warning_list.append("WARNING: Full list of {}\t{}\t{}\t{}".format(key, len(stats[key]), len(overlap_norm), overlap_norm))
+		warning_list.append("WARNING: Genes with {}. {} in raw. {} in normalized. Full list available in logging file. Preview of normalized: {}".format(key, len(stats[key]), len(overlap_norm), list(overlap_norm)[:10]))
 
 	return cxg_adata_raw_ensembl, redundant, all_remove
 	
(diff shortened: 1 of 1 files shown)


scripts/flattener.py
```diff
@@ -1647,8 +1648,10 @@ def main(mfinal_id):
 
 	# Printing out list of warnings
 	for n in warning_list:
-		print(n, end = '\n')
-		logging.warning(n)
+		if 'WARNING: Full list' not in n:
+			print(n, end = '\n')
+		if 'Full list available in logging file.' not in n:
```
jychien, line 1653: I believe `Full list available in logging file` refers to the warning that prints the first 10 drop_unmapped.

jychien: Not sure how to do this, but line 960 has also logging that needs to not print to screen, and only to file:

for key in stats:
        stats[key] = set(stats[key])
	warning_list.append("{}\t{}\t{}\t{}".format(key, len(stats[key]), len(overlap_norm), overlap_norm))

jychien:
Sample dataset to run: LATDF807HKX

Jchaffer787:
I've added code to handle the logging for "multiple_ensembl" and "multiple_symbols".
Let me know if you think there should be any adjustments on the output.
Right now the result of running LATDF807HKX is:

WARNING: Genes with multiple_ensembl. 313 in raw. 52 in normalized. Full list available in logging file. Preview of normalized: ['MEG8', 'LINC00864', 'AC006449.2', 'AC009120.4', 'AC010894.3', 'ECSCR', 'AC006369.2', 'ARHGAP11B', 'AC005253.2', 'AP003774.4']
WARNING: Genes with multiple_symbols. 22050 in raw. 2561 in normalized. Full list available in logging file. Preview of normalized: ['RP11-95D17.1', 'RP11-637A17.2', 'ATP5MC1', 'BZRAP1', 'AL139246.5', 'LINC02456', 'AC104170.1', 'AC138969.4', 'KIAA0196', 'C4orf32']

jychien:
Great, thank you! Let's merge our branches to main. And then I'll do a final check.

PR #120, opened 2024-02-09
Tools 143 improved warnings and errors for na values

Jchaffer787:
Added code to catch and output informative error in case of 'author_cell_type' column containing NaN values.

Changed warning logging so that the columns that contain NaN values are output to the command prompt screen for wrangler convenience.


jychien:
Looks good to me!

PR #122, opened 2024-02-21
Tools 123 redo atac no raw

jychien:
Also allow for dense matrix to stay as np.ndarray


jahilton (approved):

jychien:
I've added additional code that checks to see of list of gathered library objects are from multiome libraries. If so, then it only takes the appropriate library to gather metadata and is no longer considered pooled. @jahilton

PR #123, opened 2024-02-28
Tools 159 add mapping for unknown cells

Jchaffer787:
Converting 'NCIT:C17998' to 'unknown' for cell_type_ontology_term_id for both pooled and non-pooled datasets.

Also did some clean-up of calling get_value within gather_metdata. Left gather_pooled_metadata un-edited as handling is different.


jahilton (approved):
code LGTM

we should try to avoid the redundant commits (updates to HCA_data_table & curation_qa_collection) so we don't accidentally overwrite updates.
I have checked and in this case, they are consistent with main so no issue.

PR #124, opened 2024-02-28
Tools-160 add donor sex qa

corismall:


cellxgene_resources/curation_qa.ipynb
```diff
@@ -656,6 +656,133 @@
...
+   ]
+  },
+  {
+   "cell_type": "code",
+   "execution_count": null,
+   "metadata": {},
+   "outputs": [],
+   "source": [
+    "#Use raw matrix counts for sex analysis\n",
+    "raw_adata = ad.AnnData(adata.raw.X, var=adata.raw.var, obs=adata.obs)  # Create raw anndata if adata.raw exists otherwise raw_adata = adata\n",
```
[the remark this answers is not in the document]
jahilton: We'll probably also want to explore a way to do this without creating a whole other AnnData object. I can imagine some of the bigger Datasets choking at this point.

cellxgene_resources/curation_qa.ipynb
```diff
@@ -656,6 +656,133 @@
...
+    "\n",
+    "#Extract regions from adata based on feature_name \n",
+    "par_adata = raw_adata[:,raw_adata.var.feature_name.isin(par['par_genes'])]  \n",
```
jahilton, line 689: This is going to be run on pre-submission Datasets sometimes, so we can't rely on 'feature_name' being present. So probably need to rely on the Ensembl IDs in the index because that will work for pre- and post-submission Datasets.

cellxgene_resources/curation_qa.ipynb
```diff
@@ -656,6 +656,133 @@
...
+  },
+  {
+   "cell_type": "code",
+   "execution_count": null,
+   "metadata": {},
+   "outputs": [],
+   "source": [
+    "#Check for batch or library uuid fields for correct grouping\n",
+    "print(raw_adata.obs_keys)\n",
+    "library_id = 'batch'  # this variable will change depending on the dataset"
```
jahilton, line 706: We can't count on a "library" ID metadata field in all Datasets. We should try to avoid this customization.

cellxgene_resources/curation_qa.ipynb
```diff
@@ -656,6 +656,133 @@
...
+   "cell_type": "code",
+   "execution_count": null,
+   "metadata": {},
+   "outputs": [],
+   "source": [
+    "#Assign sex using 0.5 threshold\n",
+    "def assign_sex(x):\n",
+    "    if x >= 0.5:\n",
+    "        return 'Male'\n",
+    "    elif x < 0.5:\n",
```
jahilton, line 765: We talked about not having 0.5 be a strict cut-off but rather have narrower ranges for Male & Female with a gap in the middle, then keep an eye out for any that fall into the gap to see if we should adjust. I think we said keep male 0.5 and over, but female would be < 0.2

cellxgene_resources/curation_qa.ipynb
```diff
@@ -656,6 +656,133 @@
...
+   "cell_type": "code",
+   "execution_count": null,
+   "metadata": {},
+   "outputs": [],
+   "source": [
+    "#Compare each donor's labeled sex and scRNAseq-determined sex\n",
+    "check = raw_adata.obs.value_counts(['donor_id', 'sex']).reset_index().rename(columns={\"index\": \"value\", 0: \"count\"})\n",
+    "check.drop(columns=['count'],inplace=True)\n",
+    "check_donors = pd.merge(counts_avg, check, how='left', left_on='donor_id', right_on='donor_id')\n",
+    "check_donors"
```
jahilton, line 783: I can envision a Dataset with many donors, rather than report the full donor eval, you'd instead only want to report the mismatches, or the ones that are currently "unknown". So the things to act on are less likely to get buried.

jahilton (changes requested):
It seems to work, so we just want to try and make it more lightweight
I would stash the PAR_Y/non-PAR_Y lists into a single file (could be a .json that stores 2 lists, or a table that has gene in one column and PAR_Y/non-PAR_Y in another). And ref_files seems like an appropriate place to drop that.

cellxgene_resources/curation_qa.ipynb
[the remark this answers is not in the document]
corismall: > not all Datasets will have raw.X. Will need to account for those that don't (they'll have raw counts in .X)

i have this in the comment, but maybe it should be a separate line of code (commented out)?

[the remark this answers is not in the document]
jahilton: Look elsewhere in the notebook, you'll see several places where the logic is built in so the user doesn't need to pay attention `if adata.raw:`...`else:`...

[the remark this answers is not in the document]
jahilton: >We'll probably also want to explore a way to do this without creating a whole other AnnData object. I can imagine some of the bigger Datasets choking at this point.

Looking at Jenny's code from CXG-243, that code creates a new AnnData, but subset to only the genes of interest, so it's a much smaller object `adata = adata[:,adata.var['feature_id'].isin(male_genes+female_gene)]`
(but in our case, use a new variable name so that we don't overwrite the main adata object)
It looks like her ntbk may also have some ideas for more succinctly calculating the mean expression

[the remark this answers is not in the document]
corismall: Sorry if I'm not understanding correctly, but creating the smaller anndata objects is what i'm doing in these two lines [685/686](https://github.com/Lattice-Data/lattice-tools/pull/124/files#diff-c8c884b4f0b3b0bae8767087ebf2f17847330968c6337f01b163d4ff8eb9b597:~:text=%22par_adata%20%3D%20raw_adata%5B%3A%2Craw_adata,gene_ids%27%5D). 

I created a `raw_adata` variable to account for datasets that have the raw counts in `.X`, this method is faster than reusing the subsetting code for either `raw.X` or `.X`. 
```
if adata.raw:
    par_adata = adata.raw[:,adata.raw.var.index.isin(par_ids['gene_ids'])]
    nonpar_adata = adata.raw[:,adata.raw.var.index.isin(nonpar_ids['gene_ids'])] 

else:
    par_adata = adata[:,adata.var.index.isin(par_ids['gene_ids'])]  
    nonpar_adata = adata[:,adata.var.index.isin(nonpar_ids['gene_ids'])] 
```
Comparing the current code vs the above code using various datasets, the run time is longer using the code above as the datasets get larger. For example with a 5GB dataset, Elapsed time: 0.40940117835998535 seconds for current code vs Elapsed time: 1.7622332572937012 seconds for code above.

cellxgene_resources/curation_qa.ipynb
```diff
@@ -656,6 +657,129 @@
...
+   ]
+  },
+  {
+   "cell_type": "code",
+   "execution_count": null,
+   "metadata": {},
+   "outputs": [],
+   "source": [
+    "#Get PAR & Non-PAR gene_ids from ref_file\n",
+    "ref_file = 'ref_files/gtf.tsv'\n",
```
jahilton, line 750: This file does not exist
And I would choose a more descriptive variables than `ref`/`ref_file`

cellxgene_resources/ref_files/chry_regions.py
```diff
@@ -0,0 +1,573 @@
+regions = {
+    "NONPAR": [
+    "RNU6-1334P",
```
jahilton, line 3: IDs are needed for the actual QA because a pre-submission Dataset will not have symbols.
So if the symbols are listed in this file, the QA will need to do some work to get from symbol-->ID
Therefore, I think it just makes sense to store the Ensembl IDs

cellxgene_resources/curation_qa.ipynb
jahilton: Sample suggestion to simplify things...
`chry_file = 'ref_files/chry_regions.json'`
`chry_genes = json.load(open(chry_file))`
`par_ids = chry_genes['PAR']`
`nonpar_ids = chry_genes['NONPAR']`

cellxgene_resources/curation_qa.ipynb
```diff
@@ -656,6 +657,129 @@
...
+    "ref = pd.read_csv(ref_file, sep='\\t')\n",
+    "par_ids = ref[ref['gene_symbols'].isin(regions['PAR'])]\n",
+    "nonpar_ids = ref[ref['gene_symbols'].isin(regions['NONPAR'])]\n",
+    "\n",
+    "if adata.raw:\n",
+    "    #Subset PAR & Non-PAR regions from adata.raw.X\n",
+    "    par_adata = adata.raw[:,adata.raw.var.index.isin(par_ids['gene_ids'])]\n",
+    "    nonpar_adata = adata.raw[:,adata.raw.var.index.isin(nonpar_ids['gene_ids'])] \n",
+    "    \n",
+    "    #Check how many PAR & Non-PAR genes were found in dataset\n",
```
jahilton, line 760: Only the par_adata & nonpar_adata definitions are specific to the if/else logic. The rest (check_percent, region_counts, donor_df, check_donor_sex) can all be written just once after the if/else

jahilton (changes requested):
Made some suggestions based on what I could evaluate, but I believe several bugs exist that prevent it from being run so couldn't do a full evaulation.

jahilton (changes requested):

PR #128, opened 2024-03-06
removed unused dataset object query

jychien:
Was a quick update. https://lattice.atlassian.net/browse/TOOLS-168. Looks like I had some extra code floating around. Thanks for catching that, @jahilton

### DIFF
--- a/scripts/flattener.py
+++ b/scripts/flattener.py
@@ -521,10 +521,9 @@ def gather_pooled_metadata(obj_type, properties, values_to_add, objs):
 				values_to_add[key] = next(iter(value_set))
 
 
-# Gather dataset metadata for adata.uns
+# Gather matrix metadata for adata.uns
 def report_dataset(donor_objs, matrix, dataset):
 	ds_results = {}
-	ds_obj = lattice.get_object(dataset, connection)
 	for prop in dataset_metadata['final_matrix']:
 		value = get_value(matrix, prop)
 		if isinstance(value, list):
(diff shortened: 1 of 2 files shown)


jychien:
Let me know what you guys think @Jchaffer787 @jahilton. Hopefully get_report() is generic enough to be reused. I've held off on moving around any other code as I thought it would be good to not add to much to this ticket.

jahilton:
I think we want to also enable querying a report for objects if you have criteria other than a list of `@id` (so if I want all of the RawSequenceFiles with dataset=<>).
Also, if feeding a list of `@id`s then should be able to parse the type from the `@id`. 
One idea...

A more general function for querying objects...
get_report(obj_type, filters, field_lst, connection)
where filters can be anything `&dataset=<>` or `&@id=<>&@id=<>`

then have a separate function that you send a list of `@id`s to that produces obj_type & filters...
parse_ids(id_lst)
parse out the obj_type, maybe some management if the obj_types vary across `@id`s
then do the join to make the filters
returns obj_type,filters

So in practice...
obj_type, filters = parse_ids(my_list_of_@ids)
my_objs = get_report(obj_type, filters, field_lst, connection):

jahilton:
>should be able to parse the type from the @id.

Not as straightforward I was thinking
But reviewing all pairs - libraries (Library) & antibody (Antibody) are unique. 
Otherwise `obj_type = ''.join([x.capitalize() for x in at_id.split('-')]).rstrip('s')` or `obj_type = at_id.title().replace('-','').rstrip('s')` should work

jychien:
@jahilton @Jchaffer787 Let me know if this works for you guys

Jchaffer787:
Looks good to me!

PR #129, opened 2024-03-12
Tools 165 optimize code for retrieving raw matrix files

Jchaffer787:
Adding calls to lattice parse_ids and get_report in order to simplify gathering of raw matrix objects.


jychien:
Looks good to me! Thanks!

PR #132, opened 2024-03-20
Tools 175 gather geo metadaata pooled exp

jychien:
@jahilton I was able to gather the geo metadata for LATDS555LKZ. Let me know if there are any additional changes you think would be good to add or if there's anything that needs attention. I've summarized on [TOOLS ticket](https://lattice.atlassian.net/browse/TOOLS-175).

### DIFF
--- a/scripts/geo_metadata.py
+++ b/scripts/geo_metadata.py
@@ -6,7 +6,7 @@ from urllib.parse import urljoin
 import requests
 import numpy as np
 import time
-
+import gc
 
 EPILOG = '''
 Examples:
@@ -33,7 +33,7 @@ def getArgs():
     	sys.exit()
     return args
 
-library_metadata = {
+LIBRARY_METADATA = {
 	'sample': [
 		'biosample_ontology.term_name',
 		'diseases.term_name',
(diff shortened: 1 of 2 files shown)


jychien:
@jahilton I found a bug as it is not printing the md5sum for the RawMatrixFiles.

scripts/geo_metadata.py
```diff
@@ -188,35 +195,35 @@ def gather_objects(input_object, start_type=None):
...
-				unfiltered_matrices.append(obj)
-		else:
-			continue
-	return my_raw_matrices, unfiltered_matrices
 
-
-def get_dataset(dataset):
-	global geo_study
-	results = lattice.get_object(dataset, connection)
+	field_lst = RAW_MATRIX+['@id','accession', 'libraries', 'derived_from', 's3_uri', 'status']
```
jahilton, line 207: add md5sum here?

jychien: Yes! That should be it. Thanks!

jychien: Just tested it, and that was it. I'll push a commit to fix. Thanks!

jychien:
@jahilton I've updated the ordering of columns. I was going to put the constants in a separate file (ie: constants.py). But then we need a constants.py for the flattener in the same directory, so left it as is.

jahilton (approved):
LGTM

PR #133, opened 2024-03-21
CXG-233 gene mapping

jahilton:
Add documentation and script on how to evaluate gene mappings across GENCDOE versions


Jchaffer787:
LGTM!

cellxgene_resources/curation_sample_code.ipynb
```diff
@@ -849,6 +849,23 @@
...
+   ]
+  },
+  {
+   "cell_type": "code",
+   "execution_count": null,
+   "metadata": {},
+   "outputs": [],
+   "source": [
+    "v44_gene_map = json.load(open('../gene_ID_mapping/gene_map_v44.json'))\n",
+    "adata.var.rename(index=v44_gene_map, inplace=True)"
```
jychien, line 866: Jim had put in an extra check to see that the newly mapped ID is not already in the index.

jychien:
For gene_mapping.py, I looked it over but it was too complicated and couldn't following everything. But we've thoroughly gone over the output to this script, so I think that's good for the review.

cellxgene_resources/curation_sample_code.ipynb
jahilton: Good catch. Please review the added line

jychien: Looks good!

PR #134, opened 2024-03-21
Tools 177 adding outputs folder and subfolders

Jchaffer787:
Adding check for presence of and if not creation of 'outputs' folder that contains 'flattened_h5ads' and 'loggings' sub folders in order to better organize output files from flattener.

Flattened h5ad files will now be sent to 'flattened_h5ads' and logging files will be sent to 'loggings', both contained within 'outputs' folder.

Also renamed tmp_dir to mtx_dir so it's more clear what it is.

Open to any suggestions on changing naming or organization if desired.


scripts/flattener.py
```diff
@@ -1164,7 +1162,14 @@ def main(mfinal_id):
...
 	mfinal_obj = lattice.get_object(mfinal_id, connection)
-	logging.basicConfig(filename= mfinal_id + '_outfile_flattener.log', filemode='w', level=logging.INFO)
+
+	# Checking for presence / creating output folder and associated sub-folders
+	if os.path.exists(output_dir) == False:
+		os.mkdir(output_dir)
+	if os.path.exists(output_dir + '/' + fltnr_dir) == False:
+		os.mkdir(output_dir + '/' + fltnr_dir)
+
+	logging.basicConfig(filename= output_dir + '/' + fltnr_dir + '/' + mfinal_id + '_outfile_flattener.log', filemode='w', level=logging.INFO)
```
jychien, line 1172: For this filename, since there are so many variable, might be more readable as:
`logging.basicConfig(filename="{}/{}/{}_outfile_flattener.log".format(output_dir, fltnr_dir, mfinal_id), filemode='w', level=logging.INFO)`

scripts/geo_metadata.py
```diff
@@ -566,9 +568,15 @@ def main(dataset):
 	ordered_cols = [c for c in geo_samples.columns if not c.startswith(('read_','index_'))] + [c for c in geo_samples.columns if c.startswith(('read_','index_'))]
 	geo_samples = geo_samples[ordered_cols]
 
+	# Checking for presence / creating output folder and associated subfolder
+	if os.path.exists(output_dir) == False:
```
jychien, line 572: Looks like os wasn't imported to this script, so we'll need to add that to the libraries imported.

jychien:
Was thinking about this a little more. If we end up changing the directory structure of the github, we may no longer need to have the need for separate output directories. Maybe we can discuss tomorrow at our group meeting?

Jchaffer787:
After discussing with Jennifer and Erica about the directory structure, we've decided to remove the outputs subfolder for the flattener, since the flattener will likely be in it's own folder anyways in the new structure. The other scripts will keep their output subfolders for now at least.

scripts/checkfiles.py
```diff
@@ -90,6 +90,9 @@ def getArgs():
 run_id_pattern = '[a-zA-Z\d-]+'
 flowcell_pattern = '[a-zA-Z\d_-]+'
 
+output_dir = 'output'
```
jahilton, line 93: `output` vs `outputs`

jahilton (changes requested):
Likely some unification we can do with a function in lattice.py. Send the subdirectory name & the function will create `outputs` if needed, and then the subdir if needed.
Seems like the same code repeated in each script, and ensures no output/outputs differences.

Also, checkfiles has a .log file that should go into the outputs directory.

scripts/flattener.py
```diff
@@ -873,14 +873,10 @@ def set_ensembl(redundant, feature_keys):
...
-	adata.var['ensembl_ids'] = adata.var.index
-	change_list = [i for i in adata.var['ensembl_ids'] if i in gene_map.keys()]
-	change_list = [i for i in change_list if adata.var['ensembl_ids'].str.contains(gene_map[i]).any() == False]
-	for i in change_list:
-		adata.var['ensembl_ids'][np.where(adata.var['ensembl_ids'] == i)[0]] = gene_map[i]
-	adata.var.set_index('ensembl_ids',inplace=True)
+	v44_gene_map = json.load(open('gene_map/gene_map_v44.json'))
+	new_gene_map = {k:v for k,v in v44_gene_map.items() if k in adata.var.index and v not in adata.var.index}
+	test = adata.var.index
+	adata.var.rename(index=new_gene_map, inplace=True)
```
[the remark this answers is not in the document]
jahilton: Still questioning the inclusion of this update as it's not relevant to the given task or outputs re-org. Concerned that it's reverting code inadvertently (I thought the change_list calculations were necessary)

scripts/geo_metadata.py
```diff
@@ -565,10 +566,10 @@ def main(dataset):
 	geo_samples.drop(columns=collapse, inplace=True)
 	ordered_cols = [c for c in geo_samples.columns if not c.startswith(('read_','index_'))] + [c for c in geo_samples.columns if c.startswith(('read_','index_'))]
 	geo_samples = geo_samples[ordered_cols]
-
+
```
jahilton, line 569: inadvertent tab

scripts/flattener.py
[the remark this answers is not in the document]
jahilton: OK, just triple-checking that this update is intended.

jahilton (approved):

PR #136, opened 2024-03-25
Visium wrangle sample ntbk

jahilton:
establish a notebook that contributors and curators can use to curate Visium Datasets to CELLxGENE standards
Notably does not include is_primary_data & gene mapping/filtering as we typically deal with those things rather than contributors


cellxgene_resources/curation_visium.ipynb
```diff
@@ -0,0 +1,227 @@
...
+   "source": [
+    "Specify the folder that includes these files...\n",
+    "- raw_feature_bc_matrix.h5\n",
+    "- spatial/\n",
+    "  - scalefactors_json.json\n",
+    "  - tissue_hires_image.png\n",
+    "  - tissue_lowres_image.png*\n",
+    "  - tissue_positions_list.csv\n",
+    "\n",
+    "*the lowres image will be deleted but is required for `read_visium()`"
```
brian-mott, line 29: I think having this url https://www.ebi.ac.uk/biostudies/arrayexpress/studies/E-MTAB-12698 and some further comments that the output files and full res tiff are here would be helpful so someone could run the notebook on these files and see the output.

cellxgene_resources/curation_visium.ipynb
```diff
@@ -0,0 +1,227 @@
...
+  {
+   "cell_type": "code",
+   "execution_count": null,
+   "metadata": {},
+   "outputs": [],
+   "source": [
+    "folder = 'WS_PLA_S9101764'\n",
+    "\n",
+    "#REQUIRED to include empty spots, so specify raw_feature_bc_matrix.h5\n",
+    "adata = sc.read_visium(folder, count_file='raw_feature_bc_matrix.h5')"
```
brian-mott, line 41: The code block below to fill in empty 0s initially failed until I ran `adata.var_names_make_unique()`. I think further comment to clarify to run this method if the size is < 4992 and there is a warning that variable names are not unique.

cellxgene_resources/curation_visium.ipynb
```diff
@@ -0,0 +1,227 @@
...
+   ]
+  },
+  {
+   "cell_type": "code",
+   "execution_count": null,
+   "metadata": {},
+   "outputs": [],
+   "source": [
+    "#map in cell_types\n",
+    "#this example reads in a csv with barcodes in the first column and CL terms in the second column\n",
```
brian-mott, line 154: And a little more to clarify that a csv in this format can be provided to add in the cell ontology but it is not part of the output files or found in this repo

brian-mott (changes requested):
I was able to run this notebook on a few outputs from the url in this review and on another 10X mouse dataset that had less than 4992 observations. I think it would be helpful to clarify the comments in these few places and then we can see what other contributors think as they make use of this notebook

cellxgene_resources/curation_visium.ipynb
jahilton: Addressed, mostly in a new notebook header

cellxgene_resources/curation_visium.ipynb
jahilton: Good catch. Fixed by setting gene_ids as the index to start.

cellxgene_resources/curation_visium.ipynb
jahilton: Added additional comments on this. Let me know what you think

cellxgene_resources/curation_visium.ipynb
brian-mott: Looks good, looks like this now works correctly when obs is < 4992

cellxgene_resources/curation_visium.ipynb
brian-mott: This looks good too, hopefully will be clear for contributors.

brian-mott (approved):
LGTM, further clear comments and setting index resolves previous error to add 0s

PR #137, opened 2024-03-26
drop uns keys with empty values and import from unused libraries

jychien:
@Jchaffer787 I've added a check for uns keys with empty values. Also dropped import requests and urljoin.


Jchaffer787:
Looks great! Love the improved dict handling.

I also did a quick sanity check that requests and urljoin aren't being used anywhere.

jychien:
@Jchaffer787 Thank you for the double checking!

PR #140, opened 2024-04-04
update url to gencode files after cog implementation

brian-mott:
Fix url after COG update to the single-cell-curation that changed directory structure

Worked locally for me on demo and prod, Erica is testing on a local branch


Jchaffer787:
Looks good to me. Ran successfully on prod and it sounds like Erica's run on demo also ran through.

Thanks again Brian! Go ahead and merge to main!

PR #141, opened 2024-04-08
LAT-380 shrink dataset obj size

jahilton:
consequences of removing Dataset.files


jychien:
Looks good to me! I ran LATDS494QZH for geo script and LATDF477OUM, LATDF329OGL, and LATDF584NGT for flattener just in case.

PR #147, opened 2024-04-12
Tools 182 move gather functions to separate module

Jchaffer787:
Moving gather_rawmatrices, gather_objects, gather_metdata, gather_pooled_metadata, get_value to a separate gather module.

Also improved comment formatting for these functions.

Set up logger so that loggings from gather.py are sent to the same logging function as flattener.

### DIFF
--- a/scripts/flattener.py
+++ b/scripts/flattener.py
@@ -20,169 +20,7 @@ import json
 import numbers
 import flattener_mods as fm
 
-# Reference files by which the flattener will filter var features
-ref_files = {
-	'ercc':'genes_ercc.csv',
-	'human':'genes_homo_sapiens.csv',
-	'mouse':'genes_mus_musculus.csv',
-	'sars':'genes_sars_cov_2.csv'
-}
-
-# Metadata to be gathered for each object type
-cell_metadata = {
-	'donor': [
-		'donor_id',
-		'age_display',
-		'sex',
-		'ethnicity',
-		'causes_of_death.term_name',
-		'diseases.term_id',
-		'diseases.term_name',
-		'family_medical_history',
-		'living_at_sample_collection',
-		'menopausal_status',
-		'organism.taxon_id',
-		'risk_score_tyrer_cuzick_lifetime',
-		'smoker',
-		'times_pregnant'
-		],
-	'sample': [
-		'age_development_stage_redundancy',
-		'uuid',
-		'preservation_method',
-		'biosample_ontology.term_id',
-		'biosample_ontology.organ_slims',
-		'summary_development_ontology_at_collection.development_slims',
-		'summary_development_ontology_at_collection.term_id',
-		'derivation_process',
-		'diseases.term_id',
-		'diseases.term_name',
-		'disease_state',
-		'menstrual_phase_at_collection',
-		'source',
-		'summary_body_mass_index_at_collection',
-		'treatment_summary',
-		'growth_medium',
-		'genetic_modifications',
-		'@type'
-		],
-	'tissue_section': [
-		'uuid',
-		'thickness',
-		'thickness_units'
-	],
-	'suspension': [
-		'cell_depletion_factors',
-		'depleted_cell_types.term_name',
-		'derivation_process',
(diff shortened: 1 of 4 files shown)


jychien:
I am getting the following error:

> LATDF190KNY.h5ad was found locally
Traceback (most recent call last):
  File "/Users/jychien/Lattice-Data/lattice-tools/scripts/flattener.py", line 1388, in <module>
    main(args.file)
  File "/Users/jychien/Lattice-Data/lattice-tools/scripts/flattener.py", line 935, in main
    fm.gather_metdata('raw_matrix', cell_metadata['raw_matrix'], values_to_add, [mxr], connection, prop_map)
  File "/Users/jychien/Lattice-Data/lattice-tools/scripts/flattener_mods/gather.py", line 150, in gather_metdata
    value = get_value(obj,prop)
  File "/Users/jychien/Lattice-Data/lattice-tools/scripts/flattener_mods/gather.py", line 286, in get_value
    return obj.get(prop, unreported_value)
NameError: name 'unreported_value' is not defined

I think it is best to set the global contants file first before merging in these PRs, since you'll need to access these global constants. I think we can first work on TOOLS-170, TOOLS-181, and TOOLS-182 first. We can leave TOOLS-171 (global variables) for until late. After these 3 tickets, let's take a pause to see if there is more of a timeline for schema 5.1 and the visium updates.

Jchaffer787:
Had made the global constants module and set up flatttener/gather to properly import the constants. Have capitalized all constant values according to preferred standards.

jychien:
I ran through the test datasets, and LATDF477OUM came up with an error:

> % cellxgene-schema validate 63d03351-06be-478e-a0db-f7a653b6b19b_489318a0-24c3-4f5c-b105-f084ed0ea026_LATDF477OUM_v5.h5ad
Loading dependencies
Loading validator modules
Starting validation...
WARNING: Validation of raw layer was not performed due to current errors, try again after fixing current errors.
ERROR: 'unknown' in 'uns['default_embedding']' is not valid, it must be a key of 'adata.obsm'.
Validation complete in 0:00:07.187470 with status is_valid=False

Jchaffer787:
Hmm.. are you sure you don't have any local changes? I didn't get that error when I originally ran LATDF477OUM, and I just tried again with a fresh pull and it ran through smoothly as well...

jychien:
Looks like there is no field value for default_embedding for [LATDF477OUM](https://lattice-data.org/processed-matrix-files/LATDF477OUM/?format=json). I had suggested in [this comment](https://lattice.atlassian.net/browse/TOOLS-182?focusedCommentId=18288) to use gather_metdata() to get the dataset metadata. But I didn't realize that the original [report_dataset()](https://github.com/Lattice-Data/lattice-tools/blob/b4c88546a6df013999209a5d9e2337058edd358d/scripts/flattener.py#L519) would ignore fields with values == unreported_value. For most obs fields, the fields with unreported_values are dropped in [drop_cols()](https://github.com/Lattice-Data/lattice-tools/blob/b4c88546a6df013999209a5d9e2337058edd358d/scripts/flattener.py#L1100), but this does not include cleaning up uns fields. Maybe add an extra line in the code around 954 to drop default_embedding if it is unreported. Looks like that is the only field that is optional.

jychien:
Also, was thinking what the value of FLAT_VERSION was. I think initially I had used it for tracking things that made it into the flattener, but now we no longer track change logs. I am thinking, though, it would be nice to match up to CxG schema. Sometimes, I have a flattened file in my local directory, and I am wondering what version it is. What do you think if we update FLAT_VERSION to 5.1?

Other than these two things, it looks great! The commenting is very nice and clearly laid out, and it's nice being able to go to separate files when looking at code rather than one extremely long script. Thank you!

Jchaffer787:
I've pushed a commit to fix the uns['default_embedding'] issue, as well as change FLAT_VERSION to SCHEMA_VERSION

jychien:
Looks great! Really nice work getting everything so organized and clearly documented.

PR #154, opened 2024-04-25
Tools 189 compress h5ad on s3

brian-mott:
Script for batch compression of h5ad files on S3. Logging details to terminal and file to make it easy to track progress and resume if there is an interruption or error.

### DIFF
--- a/scripts/compress_s3_files.py
+++ b/scripts/compress_s3_files.py
@@ -144,7 +144,10 @@ def main(s3_uri_file):
     with open(s3_uri_file, "r") as f:
         s3_uris = [URIMetaInfo(line.strip()) for line in f]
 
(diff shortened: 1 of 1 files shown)


jychien:
Thanks, Brian, for putting this together. I'll have time early next week to look this over and test this out, if that timing works for you.

brian-mott:
Thank you Jenny. There's no rush at all since I don't think anyone actively needs to use this script. Just wanted to get this going so I don't forget about it.

jychien:
Tested the following:
- Uncompressed h5ad
- Invalid S3 uri
- File already compressed

All files were handled as expected. The logging is wonderful and is very helpful when there are many files that need to be compressed. Great work!

jychien (approved):

PR #155, opened 2024-04-29
Tools 161 update visium schema

jychien:
This PR includes both https://lattice.atlassian.net/browse/TOOLS-161 and https://lattice.atlassian.net/browse/TOOLS-188 for both Visium and Slide-seqV2 assay. 

Here are some assumptions/behaviors:
- Spatial datasets will only be of a single assay ontology
- Only visium will have spatial_s3_uri
- Will use length of unique values in library_uuid to determine if dataset is merged spatial
- Will fill in with Nan for any author obs columns if there were background spots added.
- If Slide-seq single array dataset does not have X_spatial, will error and exit.

Tested:
- LATDF645CDP: Visium with fullres
- LATDF866GTS: Merged Visium
- LATDF372LSO: Single Slide-seqV2

Wanted to get a functioning branch merged into main today, hopefully, so that Jim can use that to work on some flattener updates. The flattener should really use objects to manage the variables rather that global variables. This week we can fully test additional Visium and Slide-seqV2 datasets to make sure everything is working as expected.

### DIFF
--- a/scripts/flattener.py
+++ b/scripts/flattener.py
@@ -19,6 +19,9 @@ import matplotlib.colors as mcolors
 import json
 import numbers
 import flattener_mods as fm
+import PIL
+from PIL import Image
+PIL.Image.MAX_IMAGE_PIXELS = 933120000
 
 # Global variables
 mfinal_obj = None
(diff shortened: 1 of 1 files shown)


scripts/flattener.py
```diff
@@ -19,6 +19,9 @@
 import json
 import numbers
 import flattener_mods as fm
+import PIL
```
jahilton, line 22: Is this needed? Seems like only `Image` is used, and that is imported below
(would need to update to `Image.MAX_IMAGE_PIXELS = 933120000`

scripts/flattener.py
```diff
@@ -697,12 +680,93 @@ def check_matrix(m):
...
+# If spatial transcriptomics, add adata.uns['spatial'] fields and eventually add full res
+def process_spatial():
+	global cxg_adata_raw
+	global cxg_obs
+	global cxg_obsm
+	global cxg_uns
+	global mfinal_adata
+	global mfinal_obj
+
+	if cxg_obs['assay_ontology_term_id'].unique()[0] in ['EFO:0010961','EFO:0030062']:
```
jahilton, line 692: Is this condition needed? Seems like the `if mfinal_obj['assays'] == ['spatial transcriptomics']:` would be the gatekeeper & make this redundant

jychien: That's true. Seems like not a very useful if-statement. I think if/once we get other spatial transcript transcriptomic data, then we'll just revist.

jychien:
@jahilton I've pushed the latest commit to include dropping assay if-statement and importing PIL

jahilton (approved):
LGTM

PR #156, opened 2024-05-02
Removing global variables and replacing them with classes. Overarchin…

Jchaffer787:
Removing global variables and replacing them with classes. Overarching 'GLOBAL' class containing three separate classes 'ADATA','ADATA_PARTS', and 'MFINAL'. Each class then contains appropriate objects within.

### DIFF
--- a/scripts/flattener.py
+++ b/scripts/flattener.py
@@ -22,12 +22,7 @@ import flattener_mods as fm
 from PIL import Image
 Image.MAX_IMAGE_PIXELS = 933120000
 
-# Global variables
-mfinal_obj = None
-mfinal_adata = None
-cxg_adata = None
-cxg_adata_raw = None
-cxg_obs = None
+# Creating empty list for warnings
 warning_list = []
 
 EPILOG = '''
@@ -39,6 +34,24 @@ For more details:
 
     python %(prog)s --help
 '''
+
+class GLOBAL:
+	def __init__(self,mfinal_obj,mfinal_adata,cxg_adata,cxg_adata_raw,cxg_obs,cxg_uns,cxg_obsm):
+		self.MFINAL = MFINAL(mfinal_obj,mfinal_adata)
+		self.CXG_PARTS = CXG_PARTS(cxg_obs,cxg_obsm,cxg_uns)
+		self.CXG_ADATA = CXG_ADATA(cxg_adata,cxg_adata_raw)
+
+class MFINAL:
+	def __init__(self,mfinal_obj,mfinal_adata):
+		self.mfinal_obj, self.mfinal_adata = mfinal_obj, mfinal_adata
+
+class CXG_PARTS:
+	def __init__(self,cxg_obs,cxg_uns,cxg_obsm):
+		self.cxg_obs,self.cxg_uns,self.cxg_obsm = cxg_obs, cxg_obsm, cxg_uns
+
+class CXG_ADATA:
+	def __init__(self, cxg_adata, cxg_adata_raw):
+		self.cxg_adata, self.cxg_adata_raw = cxg_adata,cxg_adata_raw
 
 def getArgs():
     parser = argparse.ArgumentParser(
@@ -112,13 +125,12 @@ def compile_annotations(files):
 
 
 # Recreated the final matrix ids, also checking to see if '-1' was removed from original cell identifier
-def concatenate_cell_id(mxr_acc, raw_obs_names, mfinal_cells):
-	global mfinal_obj
+def concatenate_cell_id(mxr_acc, raw_obs_names, mfinal_cells, GLOBAL_OBJS):
 	new_ids = []
 	flag_removed = False
 	cell_mapping_dct = {}
-	cell_location = mfinal_obj['cell_label_location']
-	for mapping_dict in mfinal_obj['cell_label_mappings']:
+	cell_location = GLOBAL_OBJS.MFINAL.mfinal_obj['cell_label_location']
+	for mapping_dict in GLOBAL_OBJS.MFINAL.mfinal_obj['cell_label_mappings']:
 		cell_mapping_dct[mapping_dict['raw_matrix']] = mapping_dict['label']
 	for final_id in mfinal_cells:
 		if not re.search('[AGCT]+-1', final_id):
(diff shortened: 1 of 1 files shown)


jychien:
Thank, Jim! Just a few questions:
- Are classes typically all caps? From my googling, seems like constants are all caps, and classes are first letter caps. And then the object instantiated of the class is just all lowercase, like other variables
- I'm not sure if it is the all caps or maybe the extra length, but it is a little bit difficult to read. Sorry, I didn't realize this would be an issue. I know you had initially said the 3 classes were so that each could be passed separately. But now that you have the GLOBAL class, is it an option to have all the properties as properties of GLOBAL? Would it be a lot of work to drop the MFINAL, CXG_PARTS, and CXG_ADATA?

@brian-mott Did you have experience with capitalization of object classes?

scripts/flattener.py
```diff
@@ -905,11 +885,11 @@ def main(mfinal_id):
 					download_file(mxr.get('s3_uri'), fm.MTX_DIR, mxr.get('accession'))
 
 			mxr_name = '{}.h5'.format(mxr_acc) if mxr['s3_uri'].endswith('h5') else '{}.h5ad'.format(mxr_acc)
-			if mfinal_obj.get('spatial_s3_uri', None) and mfinal_obj['assays'] == ['spatial transcriptomics']:
+			if GLOBAL_OBJS.MFINAL.mfinal_obj.get('spatial_s3_uri', None) and GLOBAL_OBJS.MFINAL.mfinal_obj['assays'] == ['spatial transcriptomics']:
```
jychien, line 888: I ran the flattener for LATDF584NGT, and got the following error:

  File "/Users/jychien/Lattice-Data/lattice-tools/scripts/flattener.py", line 1264, in <module>
    main(args.file)
  File "/Users/jychien/Lattice-Data/lattice-tools/scripts/flattener.py", line 898, in main
    if 'spatial' in mfinal_adata.uns.keys():
NameError: name 'mfinal_adata' is not defined

scripts/flattener.py
```diff
@@ -333,67 +344,65 @@ def get_sex_ontology(donor_df):
...
-	global cxg_adata_raw
-	if cxg_adata_raw.shape[1] > cxg_adata.shape[1]:
-		genes_add = [x for x in cxg_adata_raw.var.index.to_list() if x not in cxg_adata.var.index.to_list()]
-		new_matrix = sparse.csr_matrix((cxg_adata.X.data, cxg_adata.X.indices, cxg_adata.X.indptr), shape = cxg_adata_raw.shape)
-		all_genes = cxg_adata.var.index.to_list()
+def add_zero(GLOBAL_OBJS):
+	if GLOBAL_OBJS.CXG_ADATA.cxg_adata_raw.shape[1] > GLOBAL_OBJS.CXG_ADATA.cxg_adata.shape[1]:
+		genes_add = [x for x in GLOBAL_OBJS.CXG_ADATA.cxg_adata_raw.var.index.to_list() if x not in GLOBAL_OBJS.CXG_ADATA.cxg_adata.var.index.to_list()]
+		new_matrix = sparse.csr_matrix((GLOBAL_OBJS.CXG_ADATA.cxg_adata.X.data, GLOBAL_OBJS.CXG_ADATA.cxg_adata.X.indices, \
+			GLOBAL_OBJS.CXG_ADATA.cxg_adata.X.indptr), shape = GLOBAL_OBJS.CXG_ADATA.cxg_adata_raw.shape)
```
jychien, line 351: Seeing how this line of code would look if we stored everything under global:
`new_matrix = sparse.csr_matrix((glob.cxg_adata.X.data, glob.cxg_adata.X.indices, glob.cxg_adata.X.indptr), shape = glob.cxg_adata_raw.shape)`

scripts/flattener.py
Jchaffer787: Oh no I must've missed running LATDF584NGT..

For the caps question, I chose to capitalize since we are trying to use these as constants, but would definitely be open to changing it to only first letter caps.

It would be possible to have all the properties under global, I just thought we wanted more organization of the properties.. it does lead to less readability though. I can go back through and change it so that they're all under global if that's whats desired.

brian-mott:
>From my googling, seems like constants are all caps, and classes are first letter caps. And then the object instantiated of the class is just all lowercase, like other variables

Yea this seems like the convention. You can also have CamelCase for class names where names with multiple words have each word capitalized and then snake_case for the variables, so lowercase and underscores between words. 

Since this is mixing a few of these concepts, maybe the class name is in CamelCase for its definition and then if its attributes are truly constant, it can be instantiated all caps? Otherwise if attributes change, they should probably be lowercase.

jychien:
Thanks @Jchaffer787 and @brian-mott!
I think the ones that we were considering that could be constants were `mfinal_adata` and `mfinal_obj`. But as Jim mentioned, they are a little different than the other constants as they need to be instantiated with an initial value determined by the ProcMatrixFile, while the other constants are the same value every time the flattener is run. 

Let's go with what Brian has suggested and use CamelCase for class names. Attributes and instances of the object should be lower case.

If there in the future could be functions specific for each of the subclasses, I think it would be worthwhile to keep the subclass objects. I think I had mentioned the possibility, but looking over the code again, each modification of each attribute is so unique, not sure if it is possible to write a general function. Thoughts?

Jchaffer787:
Yea it seems like trying to do specific functions would end up just being re-writing the entire flattener...

It's looking like the best course of action is just to get rid of the subclasses and have all the attributes be under Glob. I can get working on that today. Going to wait to hear @brian-mott 's thoughts though.

Jchaffer787:
I thought we wanted to drop the further subclasses like 'ADATA','ADATA_PARTS', and 'MFINAL'? Since they just add unwanted length to the variables when they're being called?

So if we want to have all the attributes under one class, maybe something like GlobVals?

jychien:
GlobVals is good! Yes, we're on the same page for dropping subclasses. I was thinking that all these global variables are used eventually to build an AnnData that conforms to CxG schema, so just came up with those names.

brian-mott:
GlobVals sounds pretty good, I like that name. And I think that makes sense to not get into the subclasses/nested classes. This should feel like a decent change to get better organization.

jychien:
It looking very good! Thanks so much, Jim! Ready for merge to main!

jychien (approved):

PR #157, opened 2024-05-02
Refactor cxg api usage

jahilton:
Consolidate redundant code that accesses cellxgene convenience functions and centralize to mods


cellxgene_resources/cellxgene_mods.py
```diff
@@ -42,6 +43,20 @@
...
+
+
+    from src.collection import create_collection,create_revision,get_collection,get_collections,update_collection
+    from src.dataset import create_dataset,delete_dataset,get_dataset,get_datasets,upload_datafile_from_link,upload_local_datafile
+    from src.utils.config import set_api_access_config
+
+
+    set_api_access_config(api_key_file_path)
+
+
```
brian-mott, line 59: This works correctly and is nice for a cleaner namespace. I haven't done an import into a class but it works and makes sense.

How often do you or anyone else need to set an API key for dev or staging? When I first started, I did that quite a bit as I worked on the basics, but haven't done that recently. That might require setting an instance of the class with an argument of 'dev' or 'staging' instead of this current setup with class-wide attributes

cellxgene_resources/compare_revision.ipynb
```diff
@@ -80,118 +36,92 @@
...
+    "\n",
+    "comp_df = pd.DataFrame(comp).transpose()\n",
+    "comp_df = comp_df.dropna(subset=[c for c in comp_df.columns if c != 'title'], how='all')\n",
+    "if not comp_df.empty:\n",
+    "    display(comp_df.fillna(''))\n",
+    "\n",
+    "for f in diff_props:\n",
     "    temp = comp_df[(comp_df[f + '_REV'] != comp_df[f + '_PUB']) & (comp_df[f + '_PUB'].isna() == False)]\n",
     "    for i,row in temp.iterrows():\n",
     "        print(i + '-' + f)\n",
```
brian-mott, line 127: I can't comment on the actual line, but the line with `only_in_pub = [e for e in p if e no in r]` raises a `TypeError: 'int' object is not iterable.` 

Adding something like :
`if isinstance(p, (int, float)): continue`,
will skip over the non-list values and then print out just the differences between the category columns

cellxgene_resources/compare_revision.ipynb
```diff
@@ -80,118 +36,92 @@
...
    "outputs": [],
    "source": [
-    "for f in fields:\n",
+    "for k,v in rev.items():\n",
+    "    if k not in collection.keys():\n",
+    "        if k not in should_be_absent:\n",
+    "            print('not present: ' + k)\n",
+    "    elif collection.get(k) != v and k not in should_differ_collection:\n",
+    "        if k == 'datasets':\n",
+    "            diff_props = []\n",
```
brian-mott, line 82: And related to the print out of list differences, I think making `diff_props` a set instead of a list should capture all the differences without repetition

cellxgene_resources/cellxgene_mods.py
jahilton: Actually just ran into the need to upload to dev today and hit a wall. Should definitely enable that easily. I'll make that update.

cellxgene_resources/compare_revision.ipynb
jahilton: fixed via e7323cbc3a73ce66a41960438639016ebdaced2f

cellxgene_resources/compare_revision.ipynb
jahilton: fixed via e7323cbc3a73ce66a41960438639016ebdaced2f

cellxgene_resources/cellxgene_mods.py
jahilton: @brian-mott I applied a fix in 23523fa6ed402b444395ad245887a0465c77e677 that allows for defining an environment. It seems to work, but let me know if that's what you had in mind or if you have another idea for how to implement it.

brian-mott: Looks like it works for me for both uploading datasets in `upload_local.ipynb` and pulling API information in `curation_revision.ipynb` This works for both dev and prod.

I don't think there's a need for separate instances of `CxG_API` so I think this setup without any 'self` references should be alright.

My only suggestions would be an elif clause for staging

jahilton: Good point. Added 'staging' option in 255f13d5ef814644cb9259f5d8599125f1a1d9f9

PR #164, opened 2024-06-10
drop X_spatial and any obsm that does not meet schema requirements

jychien:
I've added the modifications for dropping X_spatial and invalid obsm. 
- I modified LATDF483VSD on demo to have ‘X_umap_2’ (a single dimension numpy array) and ‘X_umap_df’ (a pandas df) added to obsm for testing if you want to give that a try.
- The spatial ProcMatrixFiles need to have the default embedding as 'X_spatial' need to have them patched to 'spatial'. Currently, they will fail CxG validation for 'default_embedding' (https://stanford-cherrylab.slack.com/archives/G01205DQWD8/p1715798337061679?thread_ts=1715787614.911629&cid=G01205DQWD8)


Jchaffer787:
Code looks good and file ran through succesfully!
LGTM!

PR #165, opened 2024-06-17
matrix chunking for evaluate_dup_counts

brian-mott:
Using `_chunk_matrix()` function from the validator in the scc repo to prevent very high memory usage with the `evaluate_dup_counts()` function.

Little to no memory increase, speed seems decent but there is the possibility for multiprocessing optimization. I've tested with sparse matrices, should check and confirm dense matrices also work.

### DIFF
--- a/cellxgene_resources/cellxgene_mods.py
+++ b/cellxgene_resources/cellxgene_mods.py
@@ -116,6 +116,7 @@ def get_adata_size(adata: ad.AnnData, show_stratified=True) -> int:
             "assay_ontology_term_id" in adata.obs.columns and
             VISIUM_ASSAY_TERM_ID in adata.obs["assay_ontology_term_id"].unique() and
             "spatial" in adata.uns.keys() and
+            "is_single" in adata.uns["spatial"] and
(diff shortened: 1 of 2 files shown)


brian-mott:
Now ready for review.

`evaluate_dup_counts()` slices into the data and indices arrays of sparse matrices to generate hashes. Matrix needs to be in sparse format. Uses almost no additional memory and very quick to hash large matrices (~10 seconds for 1.9 million cells/rows). I have tested on a variety of local h5ads, spatial and non-spatial data, and on an ec2.

`get_adata_size()` returns size of adata object and prints out individual attribute sizes. This is modified from the `adata.__sizeof__()` method; `get_adata_size()` adds ability to get sizes of raw attributes and visium image arrays in uns if they exist. Tested with a variety of h5ad files: spatial(visium and slide-seq), non-spatial, visium files curated to CXG schema, and visium h5ads created from spaceranger output.

jahilton (approved):
Looks good. Both revised functions show significant speed improvement.
Merge away!

PR #167, opened 2024-06-28
update global variable

jychien:
Quick fix to update glob variables. @Jchaffer787 I've gone through and added `glob.` to the ones Erica found when flattening LATDF393MGJ. If you could double check and scan through in case I had missed anything else. Thanks!


Jchaffer787:
LGTM.

Went through and did a quick check of the other glob variables to double check those as well.

PR #170, opened 2024-07-05
get_adata_size cleanup

brian-mott:
Fixed bug in `get_adata_size()` for visium image array check. slide seq datasets would error due to missing uns['spatial'] in second line of check.

Also moving this function to sample_curation_notebook for use when needed

### DIFF
--- a/cellxgene_resources/curation_sample_code.ipynb
+++ b/cellxgene_resources/curation_sample_code.ipynb
@@ -15,8 +15,8 @@
     "## Table of Contents\n",
     "* **CELLxGENE Revision**\n",
     " * [Remove CELLxGENE portal fields](#revision)\n",
-    "* **Adata**\n",
-    " * [Get size of adata object and attributes](#adata_size)\n",
+    "* **AnnData**\n",
+    " * [Get size of AnnData object and attributes](#adata_size)\n",
     "* **Matrix**\n",
     " * [Convert matrix to sparse](#sparsity)\n",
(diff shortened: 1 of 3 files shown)


cellxgene_resources/curation_sample_code.ipynb
```diff
@@ -15,6 +15,8 @@
     "## Table of Contents\n",
     "* **CELLxGENE Revision**\n",
     " * [Remove CELLxGENE portal fields](#revision)\n",
+    "* **Adata**\n",
```
jahilton, line 18: nit - I would header it `AnnData`

cellxgene_resources/curation_sample_code.ipynb
```diff
@@ -72,6 +74,32 @@
     "adata = revise_cxg(adata)"
    ]
   },
+  {
+   "cell_type": "markdown",
+   "metadata": {},
+   "source": [
+    "# Adata"
```
jahilton, line 81: `AnnData`

jahilton (commented):
Just changing the header. Everything else looks good.

cellxgene_resources/curation_sample_code.ipynb
brian-mott: Done, updated with latest commit

cellxgene_resources/curation_sample_code.ipynb
brian-mott: Made this change as well

jahilton (approved):
Good 2 go. Merge away

PR #174, opened 2024-07-15
hca library id dependency

jahilton:
update hca validator for req to include library_id_repository if library_id doesn't meet certain requirements


jahilton:
Got clarification that library_id_repository is optional & has no library_id dependency, so this isn't needed

PR #177, opened 2024-07-22
set non to unreported_value for cell_state

jychien:
@Jchaffer787 Here is the fix for cell_state (https://lattice.atlassian.net/browse/TOOLS-208). You can try flattening any of the Navan datasets, such as LATDF040DOH, on demo for testing a visium with actually cell_state filled out.


Jchaffer787:
Code looks fine and ran through on prod okay.

PR #181, opened 2024-07-30
Tools 201 genome annotate and fix for software field

brian-mott:
Renaming two fields:
- `mapped_reference_genome` to `reference_genome`
- `mapped_reference_annotation` to `gene_annotation_version`

`gene_annotation_version` remaps GENCODE annotation to Ensembl version. Used hardcoded map/dict in `flattener_mods/constants.py`; check in `clean_obs()` if `gene_annotation_version` column exists before mapping since not all ProcessedMatrixFiles contain annotation metadata.

Also addressing bug where `RawMatrixFile.assembly` and `RawMatrixFile.software` were filtered out in the `gather_rawmatrices()` function. Adding these fields to the local `field_lst` list now correctly gathers this metadata.

Tested using the following ProcessedMatrixFiles (various combinations of reference genomes with no annotations, one annotation, many annotations):
- `LATDF519CUZ`
- `LATDF808TGB`
- `LATDF190KNY`
- `LATDF805MJO`
- `LATDF366MLP`
- `LATDF372LSO`

Finally, typo on line 41 of `gather.py`: changed `filter_list` to `filter_lst` like in the rest of the function. This is within an if block if the object is not a RawMatrixFile; this may not have surfaced yet as a potential bug and I'm not sure if the above ProcessedMatrixFiles triggered this block. Hopefully someone else knows of a case or way to properly test this, if not, we can leave for now.

### DIFF
--- a/scripts/flattener.py
+++ b/scripts/flattener.py
@@ -539,8 +539,8 @@ def clean_obs(glob):
 		glob.cxg_obs[[i for i in glob.cxg_obs.columns.tolist() if i.startswith('family_history_')]].fillna(value='unknown')
 	
 	# map gencode to ensembl version for HCA tier 1
-	if 'mapped_reference_annotation' in glob.cxg_obs.columns:
-		glob.cxg_obs['gene_annotation_version'] = glob.cxg_obs['mapped_reference_annotation'].map(fm.GENCODE_MAP)
+	if 'gene_annotation_version' in glob.cxg_obs.columns:
+		glob.cxg_obs['gene_annotation_version'] = glob.cxg_obs['gene_annotation_version'].map(fm.GENCODE_MAP)
 
 
 # Drop any intermediate or optional fields that are all empty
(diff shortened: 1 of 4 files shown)


Jchaffer787:
reference_genome and gene_annotation_version logic looks good!
gene_annotation_version seems to be correctly mapped, or dropped if not present in the ProcessedMatrixFile.
assembly and software also seem to be correctly gathered and assigned now, good catch on that.
Overall, all LGTM!

Great catch on the filter_list typo! I suspect that was me..
Unsure about a good way to test that other than heavily modifying a file on dev... perhaps Jenny will know..

jychien:
@brian-mott Good question. I can't think of a circumstance where derived from is not RawMatrixFile. Maybe we had thought there would be a possibility of some other object that the ProcMatrixFile would be derived from. Let me ping the group and see. Or else maybe we should just remove the code.

jychien:
@brian-mott Jason is going to be updating the Lattice schema so that ProcessedMatrixFile is only derived from RawMatrixFile. So, you can go ahead and drop that part of the code in this ticket. Thank you!

brian-mott:
Removed the check and changed some formatting. My editor sometimes wants to use tabs or spaces so I reformatted one or two lines that looked wonky. Hopefully the formatting looks correct for others.

Jchaffer787 (approved):

PR #182, opened 2024-07-31
TOOLS-210 diagnosis ontology is now embedded so no need to request ontology 

jychien:
No longer need to use lattice.py to make a request, as ontology is already embedded.

Example dataset is LATDF216UIK. Flattened result is in ticket comment: https://lattice.atlassian.net/browse/TOOLS-210?focusedCommentId=19140


Jchaffer787:
LGTM!!

Jchaffer787 (approved):

PR #184, opened 2024-08-02
TOOLS-200 added required fields for Tier1 and mapped field values

jychien:
@brian-mott Here are the requirement and testing notes: https://lattice.atlassian.net/browse/TOOLS-200. I am also not sure about the correct indentations that I've made. Hopefully it adheres to python standards. Thanks!

### DIFF
--- a/scripts/flattener_mods/constants.py
+++ b/scripts/flattener_mods/constants.py
@@ -208,7 +208,7 @@ SAMPLE_COLLECTION_MAP = {
 }
 
 SAMPLE_PRESERVATION_MAP = {
-	'cyropreservation': 'frozen at -80C',
+	'cryopreservation': 'frozen at -80C',
 	'flash-freezing': 'frozen in liquid nitrogen',
 	'n/a (fresh)': 'fresh',
 	'paraffin embedding': 'paraffin block',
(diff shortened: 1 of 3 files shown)


scripts/flattener_mods/constants.py
```diff
@@ -190,3 +193,86 @@
...
+	'swab': 'brush',
+	'bronchoalveolar lavage': 'bodily fluid',
+	'aspiration': 'biopsy',
+	'density centrifugation': 'other',
+	'enzymatic digestion': 'other',
+	'cryosection': 'other'
+}
+
+SAMPLE_PRESERVATION_MAP = {
+	'cyropreservation': 'frozen at -80C',
```
brian-mott, line 211: Just randomly happened to catch this typo trying LATDF519CUZ; didn't notice with careful reading for misspellings. `cryopreservation` remained in the final object instead of remapping to `frozen at -80C`

brian-mott (changes requested):
One typo needing change, otherwise looks pretty good to me. The formatting looks correct; I suspect we might have different settings for using tabs or spaces for indents on our editors/IDEs, so might be time to further investigate and find a way to keep this consistent between us all.

scripts/flattener_mods/constants.py
jychien: Thank you for catching that! :) I've pushed a commit.

brian-mott (approved):
Looks good, maps correctly, thank you for the update!

PR #185, opened 2024-08-06
change np.bool to bool

brian-mott:
While running through standard set of ProcessedMatrixFiles to test the flattener, encountered the following error:

> Traceback (most recent call last):
>   File "/mnt/lattice-tools/scripts/flattener.py", line 1173, in <module>
>     main(args.file)
>   File "/mnt/lattice-tools/scripts/flattener.py", line 1143, in main
>     warnings = fm.copy_over_uns(glob, reserved_uns)
>                ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
>   File "/mnt/lattice-tools/scripts/flattener_mods/uns_functions.py", line 121, in copy_over_uns
>     if check_not_empty(v):
>        ^^^^^^^^^^^^^^^^^^
>   File "/mnt/lattice-tools/scripts/flattener_mods/uns_functions.py", line 86, in check_not_empty
>     and not (isinstance(value, (np.bool_, np.bool)))
>                                           ^^^^^^^
>   File "/mnt/anaconda3/envs/lattice/lib/python3.11/site-packages/numpy/__init__.py", line 324, in __getattr__
>     raise AttributeError(__former_attrs__[attr])
> AttributeError: module 'numpy' has no attribute 'bool'.
> `np.bool` was a deprecated alias for the builtin `bool`. To avoid this error in existing code, use `bool` by itself. Doing this will not modify any behavior and is safe. If you specifically wanted the numpy scalar type, use `np.bool_` here.
> The aliases was originally deprecated in NumPy 1.20; for more detail

Replacing `np.bool` with `bool`. We use numpy==1.26.4 for most python envs

Tested `LATDF742BQI` on prod and demo

PR #186, opened 2024-08-12
replace np.bool with bool

brian-mott:
Same as #185 but off of latest main changes and with a final merge into main instead of the dev branch.

Tested on demo with:

- `LATDF130FJI`
- `LATDF742BQI`


jychien:
Thanks, @brian-mott! I've ran both, and looks good. For LATDF742BQI, I got the expected: ["WARNING: The key 'log1p' has been dropped from uns due to having an empty value\n"], which is what check_not_empty() is looking for.

jychien (approved):

PR #188, opened 2024-08-22
Tools 215 fix float conversion

jychien:
Here is to fix the float to int conversion for Visium metadata. https://lattice.atlassian.net/browse/TOOLS-215

Looks like I had missed the recent merges into main, so had to merge that into my branch before merging branch into main.


jahilton (approved):
LGTM

PR #189, opened 2024-08-22
Concurrent validation for CXG

brian-mott:
Script to allow for concurrent CXG validation of AnnData h5ad files in a given directory. By default, will find h5ad files in the same directory as the script (lattice-tools/scripts).

With command line arguments:
-d / --directory DIRECTORY to specify directory to search for h5ad files. Default is script location.
-r / --revised  will only validate files with the '_revised.h5ad' suffix. Default suffix will be '.h5ad'

A process pool is created with the number of workers equal to the number of files, up to the CPU count. Validation occurs in parallel; as a worker finishes validation, the file name along with the stdout and stderr from `cellxgene-schema validate` is printed to the terminal.

Tested locally and on the EC2-flattener instance with a variety of h5ad files that both meet and don't meet the current CXG schema (5.1.0)

### DIFF
--- a/scripts/concurrent_validation.py
+++ b/scripts/concurrent_validation.py
@@ -7,11 +7,17 @@ import subprocess
 CPU_COUNT = os.cpu_count()
 DIR = os.path.dirname(os.path.realpath(__file__))
 PRINT_WIDTH = 113
-EPILOG = """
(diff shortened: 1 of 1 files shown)


jahilton (approved):
LGTM
And Jennifer provided some hands-on testing

PR #191, opened 2024-08-27
Flattener testing in parallel

brian-mott:
Basic testing setup to run the flattener in parallel, checking for any breaking changes during flattener dev work. `test_flattener.py` will create one process per inputed ProcessedMatrixFile to run the `main()` function from the flattener.

Defaults:
- Run on demo, processed matrix files from `test_processed_matrix_files.txt`

Command line arguments:

- --mode/-m [ENVIRONMENT] = demo or prod; defaults to demo if no argument provided
- --processed-matrices/-p [MATRICES] = list of ProcessedMatrixFiles to test; this list is used if --file/-f argument also provided
- --file/-f [FILE] = txt file with name of one ProcessedMatrixFile per line; defaults to `test_processed_matrix_files.txt` if no argument provided AND if no --processed_matrices/-p argument present

Final terminal printout will show either:
- {ProcessedMatrixFile}: SUCCESS   OR
- {ProcessedMatrixFile}: FAILURE {with Traceback}

After, can use `concurrent_validation.py` to quickly check that flattened files meet CXG schema.

This script should catch both handled and unhandled exceptions from the flattener.

Full run with 10 ProcessedMatrixFiles in `test_processed_matrix_files.txt` takes 6-7 min with about 50 GB of memory usage on ec2 and 30 GB of memory on local MacBook Pro. Accessing the demo on the browser seems slightly slower but this testing script does not appear to cause excessive stress on the demo or prod servers.

Tested both locally and on the flattener ec2 instance (conda env test and lattice). 

Points of concern:

- Each process needs its own Lattice connection object to query Lattice. This involves passing the object as an argument in numerous helper functions. The restructuring of the flattener with the `GlobalVals` class made this very easy, but I am still concerned I could have missed some of the changes.
- There was initially a circular import with the `download_file()` function; I moved this to `mods/gather.py`
- Possible areas with race conditions or data corruption: downloading the same RawMatrixFile, writing to the same flattened file if testing on prod and demo concurrently with the same ProcessedMatrixFiles. Reading the same file simultaneously should be safe. My biggest concern is downloading the same RawMatrixFile from two or more processes. This could be overcome by adding a function that creates a list of all h5/h5ad files needed before flattening and downloads them locally before spawning the separate worker processes.


scripts/flattener.py
```diff
@@ -1195,9 +1175,9 @@ def main(mfinal_id):
...
 
-args = getArgs()
-connection = lattice.Connection(args.mode)
-server = connection.server
 
 if __name__ == '__main__':
-    main(args.file)
+	args = getArgs()
+	connection = lattice.Connection(args.mode)
+	server = connection.server
```
jychien, line 1182: I've wondered about this line before. Is `server` being actually used anywhere?

jychien:
Amazing work, @brian-mott! Here's what I've tested:
- ran test_flattener.py with my own set of ProcMatrixFiles
- ran test_flattener.py with -p parameter
- ran flattener.py just to double check things are running smoothly

Thoughts:
- I went through and looks like the connection was all added and didn't see any that were missed
- Thank you for moving `download_file()` to gather.py, as it makes more sense to put it there so that things in uns_functions.py and flattener.py can both access it.
- So far, did not come across any issues downloading. I think the user would just need to be mindful of what they are trying to download.
- Logging is great for summarizing at end of script

Looks great! I just left one comment on that one random line of code that has been there for a while.

scripts/flattener.py
brian-mott: Not anywhere in the flattener. Git blame shows this was present since the very first flattener version, so probably just copied over from other scripts that needed to more explicitly use the `Connection.server` attribute. This line could be deleted in the flattener

`Connection.server` is the attribute to hold the environmental variable `{PROD/DEMO}_SERVER`

Some of the other scrips assign the server attribute to the `server` variable, but even then,
`connection.server` would probably be fine.

jychien: Sounds good! Let's go ahead and delete the line from the flattener. Thanks!

brian-mott: Deleted the line and quickly tested locally with `test_flattener.py` defaults and a normal flattener run with LATDF190KNY

jychien: Perfect! Thank you!

jychien (approved):

jychien:
Looks great! Thank you, @brian-mott

PR #195, opened 2024-09-13
TOOLS-218 check and convert individual raw matrix file datatype if needed

jychien:
Jennifer was getting an error when flattening a large file that had uint32 instead of float32 for raw matrix file. I've moved the check of raw matrix type to when the individual raw matrix files are being opened.


jychien:
Added a commit to update column names for the files in https://github.com/chanzuckerberg/single-cell-curation/raw/main/cellxgene_schema_cli/cellxgene_schema/gencode_files/

jahilton (approved):

PR #196, opened 2024-09-19
Tools 202 hca flag

brian-mott:
Flag for flattener when flattening towards HCA Tier 1 requirements.
Further notes here: https://lattice.atlassian.net/browse/TOOLS-202

Flag added for both flattener and flattener testing. 
Currently flag is `--hcatier1/-t`, open to different name if someone has a better idea.
Boolean flag so defaults to `False` if not set, only use flag when tier 1 requirements are needed

Use the following to check for tier 1 requirements:
`python flattener.py -m demo -f LATDF190KNY -t `
`python test_flattener.py --hcatier1`

Currently local variable in `main()` with check in an if block. 
Checks for presence of fields in TOOLS-202 and warns if not found; for fields `sample_source` and `sampled_site_condition`, will also warn if value found that is not part of enum.

Logic should work where the obs fields are kept if found in `author_columns` and present in the initial ProcessedMatrixFile.

Tested locally on demo and prod with single flattener runs and with `test_flattener.py`


jychien:
Mocked LATDF613XOQ to be Tier1 compliant on Demo.

Ran the following ProcMatrixFiles successfully without hcatier1 flag:

- LATDF190KNY: SUCCESS
- LATDF708XLX: SUCCESS
- LATDF584NGT: SUCCESS
- LATDF329OGL: SUCCESS
- LATDF613XOQ: SUCCESS

Ran the ProcMatrixFiles with hcatier1 flag, had the following appropriate warnings excepted for LATDF613XOQ on Demo:

- WARNING: 'study_pi' not present in uns for HCA Tier 1 requirements
- WARNING: 'library_preparation_batch' not present in obs for HCA Tier 1 requirements
- WARNING: 'sample_source' not present in obs for HCA Tier 1 requirements
- WARNING: 'sampled_site_condition' not present in obs for HCA Tier 1 requirements

So, looks good to me! Thank you!

There is one part of the code that I was wondering about, though. It's in add_zero(). I don't think it is technically giving an incorrect result, but was thinking maybe we should update it to using the global vars, but wanted to get your thoughts.

brian-mott:
It might be good to update to the global variables, we could do that in this ticket.

I would have to look and see what attributes do what. I'm not exactly sure what each is for. Initially for the check to see if `study_pi` was present in uns, I used `glob.cxg_uns` but that didn't seem to work compared to `glob.cxg_adata.uns`

jychien:
`glob.cxg_uns` starts out with metadata retrieved from Lattice according to DATASET_METADATA, but then things are copied over from the mfinal_adata to `glob.cxg_adata.uns` in uns_functions.py:copy_over_uns(). For spatial work, I've been using `glob.cxg_uns`, so not sure how everything has been getting in.
It looks like there is inconsistent use of `glob.cxg_uns` and `glob.cxg_adata.uns`. I will move everything over to using `glob.cxg_uns` and push a commit to this branch for your review.

brian-mott:
I'm fine either way with the fix being its own PR to merge into main (and I will review), then we merge main back into the dev branch, or we do the fix here. 

The testing features are only on the dev branch, so if there's lots of places to change to get to consistency, might be easier to do here and feel reassured with successful testing. It seems like we likely will soon have the dev branch ready for main, so hopefully if the fix is on the dev branch, it will soon be on main

jychien:
Makes sense. I agree and I think the rest of the dev branch updates should be pretty straight forward and we should be able to merge dev to main soon.

For this commit, I've made the following updates:
- When updating glob variables or during creation of glob.cxg_adata, I used the global variables.
- After glob.cxg_adata has been created, I use, for example, glob.cxg_adata.obs or glob.cxg_adata.uns, to read information from glob.cxg_adata
- Moved the call to copy_over_uns() to before glob.cxg_adata was created. This way there is less confusion and all manipulation of uns is done on glob.cxg_uns
- Changed the Tier1 checks to call on glob.cxg_adata.uns or glob.cxg_adata.obs

Let me know if this makes sense. I've re-ran the 5 test datasets, and also LATDF613XOQ on Demo as a 'valid' Tier1.

brian-mott:
Looks like all the required changes were made and this should make it easier to follow! `glob.cxg_adata` is created at line 1148 in `flattener.py` so I:

- [x] checked that functions/code that use `glob` prior to this use the specific glob attributes and not the `glob.cxg_adata` attributes
- [x] checked that functions/code after creation of `glob.cxg_adata` use its attributes
- [x] success for the 5 files mentioned above when using `test_flattener.py` on demo both with and without the hca flag
- [x] success for the standard file list with `test_flattener.py` on demo with and without the hca flag
- [x] uns is accessed correctly, print statement on line 1199 very helpful to check this 

Only change would be to remove that print statement on line 1199, otherwise good to merge!

jychien:
Thank you for catching that! I've pushed a commit to remove the print statement.

jychien (approved):

PR #198, opened 2024-09-25
update for calculate_adata_memory in sample code notebook

brian-mott:
Updating the curation_sample_code notebook to use the latest h5/h5ad memory calculation function, `calculate_adata_memory()` and removing the older function, `get_adata_size()`.  

The latest function reads a file path and returns size on disk and when fully loaded into memory. There's further instruction in the curation_sample_code notebook to enable more detailed print out of the data sizes for h5/h5ad attributes.


jahilton (approved):
LGTM

PR #199, opened 2024-09-25
Tools 205 add library id repository

Jchaffer787:
Added EGA:EGAX/SRA:SRX/ENA:ERX accessions to a new list in constants.py and added code for catching dbxrefs and filtering / joining the accessions to gather_metdata in gather.py. I’ve made a slight adjustment to drop cols to also drop optional columns with only ‘' as entries. I am open to other ideas for catching when the dbxref metadata isn’t present, but this seemed like a simple robust option that has worked in testing.

### DIFF
--- a/scripts/flattener_mods/gather.py
+++ b/scripts/flattener_mods/gather.py
@@ -197,6 +197,19 @@ def gather_metdata(obj_type, properties, values_to_add, objs, connection):
 			latkey = (obj_type + '_' + prop).replace('.', '_')
 			key = constants.PROP_MAP.get(latkey, latkey)
 			values_to_add[key] = value
+		elif prop == 'dbxrefs':
+			latkey = (obj_type + '_' + prop).replace('.', '_')
+			key = constants.PROP_MAP.get(latkey, latkey)
+			value_list = []
+			for v in value:
+				v = v.lstrip()
+				if v[:7] in constants.ACCEPTED_ACCESSIONS:
+					value_list.append(v)
+			if len(value_list) > 1:
+				values_to_add[key] = value_list[0]
+			else:
+				values_to_add[key] = ",".join(value_list)
+			# need to do same for at multiplexed.
 		else:
 			if isinstance(value, list):
 				value = ','.join(value)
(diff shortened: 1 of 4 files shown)


jychien:
Following test cases:
- **LATDF329OGL, LATDF994BQY, LATDF742BQI**: Did not have dbxrefs
- **LATDF477OUM, LATDF483VSD**: only list SRA and not GEO:GSM
- **LATDF448PLU**: mix of multiple SRA ids and single SRA (example - `SRA:SRX5914367,SRA:SRX8408726`)
- **LATDF708XLX**: ENA:ERX id (example - `ENA:ERX5202391`)

Notes:
- My understanding from @jahilton's notes from the ticket is that the first namespace should be stripped off. So, resulting names should be `SRX5914367,SRX8408726` or `ERX5202391`
- Sorry for the confusion, looks like HCA updated the field to all lowercase: library_id_repository
- I am thinking it would be good to sort the ids just to have a uniform string. Not sure, but is there any possibility of redundant dbxrefs for a single library? I am guessing not.
- I'll need to think about adding '' as a possibility for dropping the column. I want to say it's okay, but wondering why it didn't default to UNREPORTED_VALUE.
- Also, just FYI, the flattened files are all with zero genes because the fix had gone into main. But I'm just looking at obs, so should be okay.

scripts/flattener_mods/gather.py
```diff
@@ -197,6 +197,15 @@ def gather_metdata(obj_type, properties, values_to_add, objs, connection):
 			latkey = (obj_type + '_' + prop).replace('.', '_')
 			key = constants.PROP_MAP.get(latkey, latkey)
 			values_to_add[key] = value
+		elif prop == 'dbxrefs':
```
jychien, line 200: get_value() is actually returning constants.UNREPORTED_VALUE ("unknown") when there is no dbxrefs. This is the expected behavior. And what is happening is that it is not finding the constants.ACCEPTED_ACCESSIONS in the string "unknown", so it is joining an empty list, which ends up being an empty string. 
So, you'll need to add an if statement at 201:
`if value != constants.UNREPORTED_VALUE:`
And that way you won't need to make any changes to line 592.

Jchaffer787:
@jychien 
I've fixed the field to be all lowercase, and drop the first namespace.
I've added a sort to the list prior to ',' join for possibility of redundant dbxrefs, better safe than sorry.
I've removed the changes to line 592 and added the check at 201, thanks for catching that!
Have ran through the test datasets and all results are as expected.

jychien:
Looks good to me!

jychien (approved):

PR #200, opened 2024-09-27
Tools 203 required fields

brian-mott:
[More details here](https://lattice.atlassian.net/browse/TOOLS-203)

Tested on the new demo and production with and without the hca flag with the ProcessedMatrixFiles in the test file and with LATDF916QJI (recent thymus dataset with numerous enriched and depleted cell terms). LATDF483VSD (in standard test files) has a suspension with two enriched cell terms.

Currently the new columns for cell terms, `suspension_enriched_cell_terms` and `suspension_depleted_cell_terms`, are not in the optional columns list in `constants.py`. They should probably be added before the merge but I've kept them out to better check that the mapping works correctly.


jychien:
Thanks, Brian!

**manner_of_death**: 

- I am thinking that apply() may be slightly more time consuming for the larger datasets, in general, and it seems like we are getting more larger datasets over time. I have used appy() multtiple times in the flattener, and it seems to be doing okay, though, but I do reserve it for only when I I don't see any other straight forward option. One option, for manner_of_death, would be to:

> glob.cxg_adata.obs["manner_of_death"] = glob.cxg_adata.obs["donor_living_at_sample_collection"]
glob.cxg_adata.obs["manner_of_death"].replace({'True':'not applicable', 'False': 'unknown', 'na':'not applicable'}, inplace=True)

- For [prenatal](https://lattice-data.org/profiles/human_prenatal_donor) donors, donor_living_at_sample_collection = 'na'. From today's morning meeting, Corinn pointed out that these should be empty string. I've updated the jira ticket requirements.

**cell_enrichment**:  

- Thanks for leaving `suspension_enriched_cell_terms` and `suspension_depleted_cell_terms` in for testing. Before merge, I think the fields should go under COLUMNS_TO_DROP.

I ran LATDF916QJI and LATDF483VSD, as you have listed, and they looked good. Also ran LATDF013ZKP for depleted and enriched cell terms.

brian-mott:
Thanks for the comments and suggestions!

I updated the manner_of_death check to copy donor_living_at_sample_collection and use `.replace()` instead of `.apply()`. 

It looks like postnatal donors will have `True` or `False` as values and prenatal will only have `na`; the remapping via replace reflects this and also remaps empty strings to 'unknown'.
For `LATDF916QJI`, it looks like there are 'unknown' values that arrive at the point of the check and they remain as 'unknown' Is this alright or should we remap to something else or is there some cleanup step that is replacing a default value with unknown?

For cell_enrichment, it still uses `.apply()`. I tried some sequential additions to the cell_enrichment column but this seemed less clear than the current approach. Open to another way if that would be easier/faster.

jychien:
Looks good to me! I ran the full dataset and they all ran successfully. 

> For LATDF916QJI, it looks like there are 'unknown' values that arrive at the point of the check and they remain as 'unknown' Is this alright or should we remap to something else or is there some cleanup step that is replacing a default value with unknown?

Looks like for the one that are 'unknown' values are pediatric samples and unknown for 'donor_living_at_sample_collection'. It's a little ambiguous. 'not applicable' does not make sense. 'unknown' is still a little ambiguous, because Tier1 schema means 'the cause of death is unknown", but I don't think there is another option.

> For cell_enrichment, it still uses .apply(). I tried some sequential additions to the cell_enrichment column but this seemed less clear than the current approach. Open to another way if that would be easier/faster.

Sounds good!

jychien (approved):

PR #203, opened 2024-10-07
Tools 206 rename or add remaining tier1 optional fields

Jchaffer787:
**library_starting_quanitity + _units changed to cell_number_loaded + _units**
- Changed values in prop_map and optional_columns in constants.py
- Tested on LATDF190KNY  

**suspension_percent_cell_viability changed to cell_viability_percentage**
- Added mapping to PROP_MAP in constants.py
- Replaced suspension_percent_cell viability with cell_viability_percentage in OPTIONAL_COLUMNS in constants.py 
- Tested on LATDF879UES

**tissue.date_obtained changed to sample_collection_year**
- Added date_obtained to sample part of cell_metadata
- Added map to ‘sample_collection_year’ to prop_map
- Added ‘sample_collection_year’ to optional_columns
- Added code to gather_metdata in gather.py to strip un-needed day and month info if present. 
- Tested on LATDF519CUZ (with dashes) and LATDF419ACG (without dashes)

**Adding intronic_reads_counted**
- added intronic_reads_counted to CELL_METADATA
- Added mapping to intronic_reads_counted to PROP_MAP in constants.py
- Added ‘intronic_reads_counted’ to gather_rawmatrices function in gather.py
- Added handling to gather_metdata to convert boolean. 
- Tested on LATDF190KNY

Note: MOVED dbxrefs in prop_map to be with other library terms, just for organizations sake. This might be why it cannot automatically merge?

### DIFF
--- a/scripts/flattener_mods/gather.py
+++ b/scripts/flattener_mods/gather.py
@@ -231,9 +231,9 @@ def gather_metdata(obj_type, properties, values_to_add, objs, connection):
 		elif prop == 'date_obtained':
 			if value != constants.UNREPORTED_VALUE:
 				value = value.split('-')[0]
-				latkey = (obj_type + '_' + prop).replace('.', '_')
-				key = constants.PROP_MAP.get(latkey, latkey)
-				values_to_add[key] = value
+			latkey = (obj_type + '_' + prop).replace('.', '_')
+			key = constants.PROP_MAP.get(latkey, latkey)
+			values_to_add[key] = value
 		elif prop == 'intronic_reads_counted':
 			if value != constants.UNREPORTED_VALUE:
 				value = 'yes' if value else 'no'
@@ -318,6 +318,9 @@ def gather_pooled_metadata(obj_type, properties, values_to_add, objs, connection
 					if v == 'NCIT:C17998':
 						v = 'unknown'
 					value.append(v)
+				if prop == 'date_obtained':
+					if v != constants.UNREPORTED_VALUE:
+						v = v.split('-')[0]
 				if isinstance(v, list):
 					value.extend(v)
 				else:
--- a/scripts/flattener_mods/gather.py
+++ b/scripts/flattener_mods/gather.py
@@ -317,7 +317,6 @@ def gather_pooled_metadata(obj_type, properties, values_to_add, objs, connection
 				if prop == 'cell_ontology.term_id':
 					if v == 'NCIT:C17998':
 						v = 'unknown'
-					value.append(v)
 				if prop == 'date_obtained':
 					if v != constants.UNREPORTED_VALUE:
 						v = v.split('-')[0]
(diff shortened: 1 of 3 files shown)


jychien:
Thanks! Just a few things that I saw when running through the standard set of datasets:

**library_starting_quanitity + _units changed to cell_number_loaded + _units**

- LATDF190KNY: cell_number_loaded should have the units (cell_number_loaded_units ) concatenated to the numeric field (cell_number_loaded). So, for this sample dataset, it should be, for example, `6000 cells`. Field names should be updated in https://github.com/Lattice-Data/lattice-tools/blob/3d07023fd8296ba3b32084c2ad6eb4f031d586ec/scripts/flattener.py#L511
- LATDF519CUZ: From library [f36249df-5147-458a-965b-5aed84df6129](https://lattice-data.org/libraries/LATLB332MWF/?format=json), it has starting quantity of 110000. The number is pretty high and it is not a multiplexed experiment. I wonder if our fields are capturing the same information. (LATDF708XLX has 1000000)
- LATDF448PLU: This dataset has a mix of known and unknown starting quantities. Is this okay?
- LATDF366MLP: Just FYI, this has a range 6000-8000. I am thinking this should be okay, but might be good to double check.


**tissue.date_obtained changed to sample_collection_year**
- LATDF477OUM: Is returning 'unknown'. Not sure why it is not being dropped.

Jchaffer787:
I've been investigating LATDF477OUM returning 'unknown' for sample_collection_year.

It looks like it was simple issue of the handing I had in gather_metdata not actually filling in the value if its the UNREPORTED_VALUE, resulting in a nan. On the other hand, gather_pooled_metdata, which I've now realized is also called, returns the expected 'unknown'. This results in the unique values in the obs column being > 1, so the column isn't dropped by drop_cols.

I've fixed the issue with gather_metdata, and added some handling for gather_pooled_metadata so that if there's ever another pooled case with the years it will be handled appropriately. While looking at this section of the code, I'm realizing the handling of 'cell_ontology.term_id' in gather_pooled_metadata might be incorrect. Currently it looks like it will incorrectly append the value twice to the list. Resulting in duplicate values. Would be interesting to check a test case for this. I will add a comment to the specific line I'm looking at.

scripts/flattener_mods/gather.py
```diff
@@ -305,6 +318,9 @@ def gather_pooled_metadata(obj_type, properties, values_to_add, objs, connection
 					if v == 'NCIT:C17998':
 						v = 'unknown'
 					value.append(v)
```
Jchaffer787, line 320: This line might need to be removed? It appends v to value, and then later on that same v is appended to value again in the else statement.

jychien: Yes! Thank you for catching that! Let's drop the append()

Jchaffer787:
@jychien I thought we wanted to change the handling of the cells/ml to convert them to cells/ul, are we not going with that idea?

jychien:
My understanding was that we would eventualy Lattice schema, drop cells/ml as an enum for units, and then update the values. Could you double check with the group?

jychien (approved):

PR #204, opened 2024-10-10
Tools 219 comments for flattener testing

brian-mott:
Add support for comments within the flattener testing txt input file.

Before only one ProcessedMatrixFile per line, which makes it hard to remember why each file was chosen for testing. Now the standard processed matrix files in `test_processed_matrix_files.txt` are commented with the specifics for why they are used for testing.

Txt input files support comments that start with '#'. Anything to the right of the # will be ignored, so full line comments and inline comments are possible.

Also fixed syntax error in constants with this PR


jychien (approved):

jychien:
Looks good to me! Thanks!

PR #208, opened 2024-10-16
TOOLS-227 convert boolean to string if background spots added

jychien:
If there are <4992 spots, check to see that any boolean glob.cxg_obs columns are converted to string. Example dataset is LATDF568AAQ.


brian-mott (approved):
Looks good! 

Tested on prod with LATDF568AAQ that brought to light initial bug and also with LATDF584NGT (Visium test file with < 4992 spots).  For LATDF584NGT, I locally changed author column to bool all `True`, all `False`, and mix of `True` and `False`. Both flattened correctly and the flattened column now is categorical with string 'True/False' and string 'nan' for filled in background spots.

Thanks Jenny!

jychien:
Thank you for the thorough review, @brian-mott ! It's been merged to main.

PR #210, opened 2024-10-22
TOOLS-228 check for empty obsm and instantiate with copy from raw for visium

jychien:
Currently, flattener is producing h5ad without any embeddings if Visium ProcMatrixFile are submitted without any embeddings. The flattener should be adding the ‘spatial’ embedding. Example: LATDF704EYC

There should be a warning or check if there are no embeddings. Remove old logging that is no longer relevant about ‘default_visualization’.


brian-mott (approved):
Looks good!

Tested with LATDF704EYC, LATDF426GQK (looks like another spatial h5ad split from the integrated object for the same collection) and LATDF584NGT (standard test file for visium datasets). All flattened and were validated True with the CXG validator.

If I remove lines 654-657, the files will not flatten and the flattener errors and exits with the new error check looking for at least one embedding.

PR #212, opened 2024-10-26
Small things oct2024

jahilton:
Changed the barcode reader so that it can be run if assay_ontology_term_id isn't yet defined (eg can be run per sample, donor_id or any other obs column)
Removed a 'future.no_silent_downcasting' that is causing errors when running in the hub. Doesn't seem to be needed anymore.


brian-mott (approved):
Looks good to me, works without error on Jupyter hub and works with datasets that do not yet have `assay_ontology_term_id` defined.

PR #215, opened 2024-11-12
Primary portion fix and exclude deprecated dev stage terms

brian-mott:
See TOOLS-232 for further info.

Rearranged ordering of code for primary.portion that caused this error while flattening a dataset with an author column with primary_portion:
`ERROR: 'obs_field' value 'study_name' not found in cxg_obs columns`

Also added new parameter to query to exclude deprecated dev terms for pooled donor metadata.

Tested on ec2 instance on prod with the normal test files and these 3 processed matrix file that contain author columns with primary_portion:

- LATDF762QWB
- LATDF065UGW
 -LATDF879UES

Jennifer also successfully flattened the initial file with this fix: LATDF343OVF

### DIFF
--- a/scripts/flattener.py
+++ b/scripts/flattener.py
@@ -1225,19 +1225,6 @@ def main(mfinal_id, connection, hcatier1):
 	del cxg_adata_lst
 	gc.collect()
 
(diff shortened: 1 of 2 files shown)


jychien (approved):

jychien:
Looks good to me! Thanks, Brian!

PR #216, opened 2024-11-14
Tools 234 drop pooled library repo prefix

jychien:
Tested the following:
- The test_tier1_matrix_files.txt testset
- **LATDF910UFX**: Chen dataset with the pooled SRX, and the GSM IDs are dropped
- **LATDF469ZOO**: Has EGAX has a dbxref


brian-mott (approved):
Looks good!
Tested same files on prod and I don't see any GSE/GSM accessions in library_id_repository. The pooled and non-pooled donors have the appropriate values, and the accessions match up with the library_id/uuid in Lattice.

PR #217, opened 2024-11-18
Cxg 5 2 qa testing

brian-mott:
Need to actually merge the testing branch from the prior migration. After merge, can think about file reorganization to prevent duplication of the h5ad test fixtures and keep imports across files as simple as possible.


jahilton (approved):
LGTM

PR #218, opened 2024-11-18
concurrent validation txt file input

brian-mott:
Improve workflow for flattener testing and validation. Now a txt file with one processed matrix file per line can be fed as input to test_flattener.py and concurrent_validation.py to both flatten and then validate the output respectfully. 

concurrent_validation.py now can take a txt file input in the same format as test_flattener.py. This update searches for h5ad files that contain the processed matrix file accession within the file name and then will only validate those flattened files, instead of all the flattened files in lattice-tools/scripts.

If there is a common prefix/suffix in h5ad files used in other directories, this update can also help to better filter down files to concurrently validate.

Tested with multiple files in scripts, multiple files in other directories, both with the '_revised.h5ad' suffix and without.

### DIFF
--- a/scripts/concurrent_validation.py
+++ b/scripts/concurrent_validation.py
@@ -14,10 +14,12 @@ Script to run CXG validation in parallel.
 By default will collect h5ad files from lattice-tools/scripts directory
 Use argument -d --directory to specify other location
 Use argument -r --revised to only collect h5ad files with the '_revised.h5ad' suffix
+Use argument -t --testfile to select h5ads from a test_flattener input txt file
 
 Examples:
     python {SCRIPT_NAME} -d /mnt/test_files -r
     python {SCRIPT_NAME} --revised --directory /Users/me/Documents/curation/files
+    python {SCRIPT_NAME} --testfile test_processed_matrix_files.txt
 """
 
 def getArgs():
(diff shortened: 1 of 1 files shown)


jychien:
This is great! Very useful implementation of concurrent_validation.py. Things tested:
- ran concurrent_validation.py after test_flattener.py with same txt file
- ran concurrent_validation.py with --revised option with directory that had h5ad file with '_revised.h5ad' in name
- ran concurrent_validation.py with --revised option with directory that did not have h5ad with '_revised.h5ad' in name, and got logging:

  > Found 0 REVISED h5ad(s) in /home/jovyan/curation/cxg-535/ to validate
  =================================================================================================================
  > 
  >   File "/home/jovyan/lattice-tools/scripts/concurrent_validation.py", line 105, in <module>
  >     validate_all_files(files)
  >   File "/home/jovyan/lattice-tools/scripts/concurrent_validation.py", line 97, in validate_all_files
  >     with multiprocessing.Pool(processes=workers) as pool:
  >          ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  >   File "/opt/conda/lib/python3.11/multiprocessing/context.py", line 119, in Pool
  >     return Pool(processes, initializer, initargs, maxtasksperchild,
  >            ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  >   File "/opt/conda/lib/python3.11/multiprocessing/pool.py", line 205, in __init__
  >     raise ValueError("Number of processes must be at least 1")
  > ValueError: Number of processes must be at least 1

brian-mott:
I can catch that exception to make the error message a little cleaner, but it didn't seem that hard to figure out the input should be different to find files to validate

jychien:
Yes, I was thinking the same thing. The ValueError is pretty straight forward and easy to understand, plus the first line in the logging shows that there aren't any _revised files in the directory. 
Looks good to me!

PR #223, opened 2024-12-05
Tools 216 cog pooled metadata

brian-mott:
See [TOOLS-216](https://lattice.atlassian.net/browse/TOOLS-216) for more details.

Fix for finding best common ontology term for pooled donors in the flattener. Makes use of COG, `find_common_ontology_term()` should work for any ontology used by CXG and COG.

Tested on these datasets on prod:

- LATDF065UGW: Humphreys dataset with non-pooled diabetic kidney donors with lots of donor diseases. experimental_variable_disease set to diabetes term
- LATDF826ZFN: other Sanes eye dataset with pooled donors, sex set to unknown and pooled donors get ‘seventh decade’ or ‘prime adult stage’ dev term depending on pooling
- LATDF477OUM: Sanes dataset with different pooling than above, ‘adult stage’ dev term for pooled donors
- LATDF216UIK: standard test file with pooled experiments. Can test on demo with pooled disease, should error and exit
- LATDF910UFX: subset of retina dataset with several pooled donors. Initially wrong linkage with donors pooled at tissue instead of further downstream; should now flatten correctly with appropriate pooled terms

Two notes/questions:

- Currently assert in middle of `find_common_ontology_term()` that will fire if Lattice returns deprecated terms. This will cause the flattener to stop and error. With printout/logging of which pooled donor and terms were inputed, this seems alright to me to make sure Lattice is using current terms and/or it will be easy to fix errors in Lattice
- `find_common_ontology_term()` might be helpful in other curation tasks. Should this live somewhere in cellxgene_resources or cellxgene_mods to make it easier to import in curation workflows?

Also merge conflict with main that we can go over together for merge conflict practice.


Jchaffer787:
Testing results:
Ran files and checked h5ads.

**LATDF065UGW:** Appropriate handling. disease_ontology is either normal or appropriate diabetes mondo term.

**LATDF826ZFN:** Took a peek at the 'reported_diseases' column for this, and some of the values are '[metastatic melanoma]', are we expecting these brackets here? In terms of handling the dev stage pooling, looks perfect. The warning printouts are very nice and easy to follow.

**LATDF477OUM:** Also looks good regarding the development_stage mapping.

**LATDF216UIK:** Correctly errors out on demo with informative message.

**LATDF910UFX:** Looks good! Double checked the find_common_ontology_term() results too, looks good.


As for your notes/questions:
I agree that the current assert printout / logging of the deprecated looks acceptable. Maybe run it by Jennifer and Erica to make sure they agree.

Yes, I think it would be helpful to put it into cellxgene_mods to make it easily accessible for notebook curation etc. good idea!

Overall, LGTM! Only thing to note is the random brackets in 'reported_diseases' for LATDF826ZFN. I might just be forgetting how we wanted to handle that though.

PR #224, opened 2024-12-11
TOOLS-230 barcode eval update

jahilton:


corismall:
Works well for me! Helped narrow down the barcode issue I was seeing in Andrew's dataset to a few problematic donors.

jychien:
Tested the following:

- Used a standard dataset with mixed assays. It looks goods and the output barcode_df has a helpful output to narrow down any discrepancies.
- Ran the 24bp Flex dataset and evaluate_10x_barcodes() reported all 24nt for all cells
- Dropped the last 8bp of the 24bp Flex dataset, resulting in the 16bp cell barcodes for Flex. evaluate_10x_barcodes() reported the following.
Screenshot 2024-12-12 at 4 29 30 PM

Looks good to me! Thanks!

PR #227, opened 2024-12-18
Tools 235 move loggings warnings to glob

Jchaffer787:
Added warning_list to global. Added passing global to any functions that had warnings but didn't already take global.
Tested on standard test datasets as well as LATDF700WAP and LATDF695OQI, which were test datasets for when we made changes to the warnings system in an earlier ticket.

All printout warnings look good and logger warnings look good as well.

### DIFF
--- a/scripts/flattener.py
+++ b/scripts/flattener.py
@@ -20,9 +20,6 @@ import numbers
 import squidpy as sq
 import flattener_mods as fm
 
-# Creating empty list for warnings
-warning_list = []
-
 EPILOG = '''
 Examples:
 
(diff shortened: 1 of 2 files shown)


jychien:
I ran the standard test datasets and LATDF483VSD and LATDF216UIK had the following error:

  File "/home/jovyan/lattice-tools/scripts/test_flattener.py", line 83, in flatten
    main(file, connection, args.hcatier1)
    ^^^^^^^^^^^^^^^^^
  File "/home/jovyan/lattice-tools/scripts/flattener.py", line 1245, in main
    clean_obs(glob)
         ^^^^^^^^^^^
  File "/home/jovyan/lattice-tools/scripts/flattener.py", line 499, in clean_obs
    warning_list.append("WARNING: author_column not in final matrix: {}".format(author_col))
       ^^^^^^^^^^^^^^^^^
NameError: name 'warning_list' is not defined

Looks like there is a reference `warning_list` in `clean_obs()` that should be updated to `glob.warnings`

Jchaffer787:
Interestingly, when I run those datasets (on prod) they run through fine with no error, but there is a missed warning_list in clean_obs which I've replaced. Thanks for catching that!

jychien:
Looks good to me! Thanks, Jim!

PR #228, opened 2025-01-11
TOOLS-238 for fields that have a value of an empty list, default value to unreported value

jychien:
This fix allows fields, including 'sequencing_platform', to have empty lists as their value.

- Ran standard test set and looked at sequencing_platform, and returned valid values. 
- The current code on main will fail on DEMO for LATDF684CEO, which was the dataset that Jennifer was working on when she ran into an error. Current code on this branch will flattener correctly to sequencing_platform to "unknown"
- Also removed code for when sequencing_platform was being filled out manually rather than it's current implementation through checkfiles

### DIFF
--- a/scripts/flattener.py
+++ b/scripts/flattener.py
@@ -581,9 +581,6 @@ def clean_obs(glob):
 
 # Drop any intermediate or optional fields that are all empty
 def drop_cols(celltype_col, glob):
-	if 'sequencing_platform' in glob.cxg_obs.columns:
-		if glob.cxg_obs['sequencing_platform'].isnull().values.any():
-			glob.cxg_obs['sequencing_platform'].fillna(fm.UNREPORTED_VALUE, inplace=True)
 	for col in fm.OPTIONAL_COLUMNS:
 		if col in glob.cxg_obs.columns.to_list():
 			col_content = glob.cxg_obs[col].unique()
--- a/scripts/flattener_mods/gather.py
+++ b/scripts/flattener_mods/gather.py
@@ -255,7 +255,10 @@ def gather_metdata(obj_type, properties, values_to_add, objs, connection):
 				values_to_add[key] = value
 		else:
 			if isinstance(value, list):
-				value = ','.join(value)
+				if len(value) == 0:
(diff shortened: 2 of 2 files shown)


brian-mott (approved):
Looks good to me!

Tested standard test files on prod and demo; all successfully flattened and were CXG valid. LATDF684CEO on demo now flattens with these changes, fails when using the current flattener on main. All flattened files did not contain an empty list for a value in `sequencing_platform`; had 'unknown' if `sequencing_platform` didn't have a value.

Jchaffer787 (approved):
Ran through and checked 'sequencing platform' values for test datasets.

This is almost certainly on my end, but I keep getting failure for LATDF216UIK:
ERROR: Pooled disease ontology for 'pooled [D9,D7,D8,D10,D6]' contains multiple values {'unknown', 'MONDO:0007254'}

Is this expected? I made sure to redownload raw matrix files, and still get the same result.

@brian-mott We're you able to run LATDF684CEO on local? It causes a memory crash for me once it gets to LATDF128NOF.h5

Code itself looks good though.

brian-mott:
@Jchaffer787 For your questions:

- For LATDF216UIK on demo, with [TOOLS-216](https://lattice.atlassian.net/browse/TOOLS-216) we changed one of the pooled donors to have multiple diseases to make sure we error and exit in this situation, so that's the expected result. Might be good to change that back to normal now that TOOLS-216 is merged and complete
- I was able to run LATDF684CEO locally on demo through `test_flattener.py`; I just added it to the end of `test_processed_matrix_files.txt` It did take the longest out of all the standard test files and its memory usage got up to at least ~30 GB, though I was only glancing at btop here and there; the actual memory peak might have been much higher.

Jchaffer787:
Thanks @brian-mott just wanted to double check.

I was able to run LATDF684CEO locally on demo after restarting my computer, in case anybody ever encounters a similar issue. Also seeing the 'unknown' for no value.

@jychien Looks like were all good to merge!

jychien:
@brian-mott @Jchaffer787 I've updated LATDF216UIK donor "D10" on Demo back to what it was as on prod.

PR #229, opened 2025-01-13
TOOLS-243: Automate pathfinding for CXG repos and API keys

brian-mott:
Currently, `cellxgene_mods.py` will look to `~/GitClones/CZI/` and  `~/Documents/keys/cxg-api-key.txt` to find local copies of these assets. Through time and with the use of cloud resources like JupyterHub and EC2 instances, this directory structure varies across machines. It becomes bothersome to manually update these paths when needing to switch branches, remotely pull new commits, etc. 

This implementation automates the process by looking in several likely places for the files and using `Path()` from the included `pathlib` library/module. These places cover all locations used by Lattice as polled last week, save locations that are directly within a local repository. There is a small risk to accidentally expose the keys through inadvertently adding them to a commit; hopefully with this update, the keys can be placed in one location outside of a repo and paths will not need to be manually changed.

Open to other suggestions or possibly using glob/recursive glob for a cleaner implementation, but that seemed to add anywhere from 10-20 seconds of time to search, where this setup is about instantaneous. 

Tested on local MacBook Pro, flattener ec2 instance, and JupyterHub, with various paths used.

### DIFF
--- a/README.md
+++ b/README.md
@@ -59,7 +59,16 @@ Compiles studies from CELLxGENE, HCA Data Portal, HCA Publications, and Bionetwo
 * **upload_local.ipynb**
 Submitting local files to CELLxGENE<br>
 Please note: <br>
-    This script utilizes the [single-cell-curation](https://github.com/chanzuckerberg/single-cell-curation/tree/main) repo which should be cloned to the following directory `~/GitClones/CZI/` and CXG API keys should be stored in `~/Documents/keys/cxg-api-key.txt`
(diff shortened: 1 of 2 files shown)


jychien:
Thanks for this! I currently have:

`single-cell-curation`: 
- locally and Hub:  `~/`

CXG API keys:
- locally: `~/Documents/keys/`
- Hub: `~/`

Ran the import lines from `curation_qa_collection.ipynb` and `curation_qa.ipynb` and both were able to find the `single-cell-curation` and keys appropriately.

jahilton (approved):
LGTM & worked in the hub & locally

PR #234, opened 2025-02-06
TOOLS-249 add python script to generate upload presigned url

jychien:
generate_presigned_url.py takes the following inputs:

    -f : file to upload
    
    -b : aws s3 bucket name
    
    -n : path and name of file that will appear in S3 (optional, defaults to directly in bucket with same file name as input file)
    
    -t : length of time that URL will be valid for (optional, defaults to 1 hour)

This presigned URL will overwrite if there is an existing file already on S3. Jennifer had said that this was fine, since many times they are regenerating a curated ProcMatrixFile and will need to overwrite. Script will display a WARNING.

### DIFF
--- a/scripts/generate_presigned_url.py
+++ b/scripts/generate_presigned_url.py
@@ -1,0 +1,77 @@
+import argparse
+import sys
+import boto3
+import botocore
+
+
+EPILOG = f"""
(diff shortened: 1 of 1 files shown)


Jchaffer787 (approved):
Code looks good, and I was able to generate a url and upload to a test bucket.

LGTM!

PR #235, opened 2025-02-06
Tools 239 check presence of prefix suffix in procmatrixfile

Jchaffer787:
Added prefixes_suffixes_presence_and_duplicate_check method to flattener.py that checks for presence of all prefixes/suffixes in ProcMatrixFile, and gives appropriate error if any are not present or redundant (duplicate).

Added fix to issue with prefixes being substrings of other prefixes.

Tested by modifying LATDF645IBZ on demo, changed mapping labels to be duplicate, not in ProcMatrixFile.

Also ran test_flattener.py, but it's coming back with failures that don't seem related to these changes:
```
LATDF584NGT: FAILURE
anndata._io.utils.AnnDataReadError: Above error raised while reading key '/obs/__categories/orig.ident' of type <class 'h5py._hl.dataset.Dataset'> from /.
LATDF329OGL: FAILURE
anndata._io.utils.AnnDataReadError: Above error raised while reading key '/obs/__categories/celltype' of type <class 'h5py._hl.dataset.Dataset'> from /.

```

### DIFF
--- a/scripts/flattener.py
+++ b/scripts/flattener.py
@@ -799,7 +799,6 @@ def hcatier1_check(glob):
 		glob.cxg_obs['library_sequencing_run'] = 'unknown'
 
 # Sanity check that all prefixes/suffixes are actually in ProcMatrixFile
-# NEED TO ADD CHECK FOR NO REDUNDANCY IN PREFIXES/SUFFIXES 
 def prefixes_suffixes_presence_and_duplicate_check(glob):
 	mfinal_cell_identifiers = glob.mfinal_adata.obs.index.to_list()
 	if glob.mfinal_obj.get('cell_label_location') == 'prefix':
--- a/scripts/flattener.py
+++ b/scripts/flattener.py
@@ -798,6 +798,32 @@ def hcatier1_check(glob):
 	else:
 		glob.cxg_obs['library_sequencing_run'] = 'unknown'
 
+# Sanity check that all prefixes/suffixes are actually in ProcMatrixFile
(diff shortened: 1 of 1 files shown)


brian-mott:
The test failures might be related to AnnData version. If you are using AnnData >= 0.11, that could break some things. I have my Lattice env I use for the flattener that has Anndata at 0.10.9. The CXG 5.3 validator is using AnnData >= 0.11.

Could be other version inconsistencies, but running `test_flattener.py` on demo with my normal Lattice env worked.

scripts/flattener.py
```diff
@@ -798,6 +798,31 @@ def map_cell_enrichment(row):
 	else:
 		glob.cxg_obs['library_sequencing_run'] = 'unknown'
 
+# Sanity check that all prefixes/suffixes are actually in ProcMatrixFile
+def prefixes_suffixes_presence_and_duplicate_check(glob):
+	mfinal_cell_identifiers = glob.mfinal_adata.obs.index.to_list()
```
jychien, line 803: `mfinal_cell_identifiers` outside of this function is defined on line 878. Might be good for larger datasets to pass the previously defined `mfinal_cell_identifiers` to this function so that the list doesn't have to be created again.

jychien:
Looks good, Jim! Thank you! I just have the one small comment.

Jchaffer787:
@jychien Great call! I've added mfinal_cell_identifiers to the inputs of the function! I also added a print out of the duplicate prefixes/suffixes in the error so it'll be easier for users to find.

jychien (approved):

jychien:
Looks good! Thanks for adding the logging, I think that will be helpful!

PR #237, opened 2025-03-24
Tools 250 curation qa ntbok 53

jychien:
Updated curation notebooks for the following:

- establish reasonable thresholds for gene counts for each organism & include those checks
- evaluate_sparsity() - looks like the check is correct (anything 50% sparse or more has to be sparse csr_matrix) but should be ERROR now instead of WARNING
- check that the gene IDs are from the organism in obs - ERROR for any with mixed organism in obs
- curation_visium.ipynb:
  - addition to adata.obs['assay_ontology_term_id'] to list out other possible assays for V2
  - update if adata.obs.shape[0] < 4992 to be assay specific

Current logic:
- if there is more than a single organism according to either the indices in var or organism_ontology_term_id in obs, `evaluate_var_df()` will report the error and then not go through with with number of genes. Please let me know if this behavior should be updated.
- the assumption is that adata.var and adata.raw.var will be matching, so looks only at adata.var

Things checked:
- Ran notebook using a correct AnnData, resulted in `GOOD:Single organism found: {'NCBITaxon:9606'}` and `GOOD:36387 genes out of possible 39583 present`
- Ran notebook using AnnData with < threshold genes for human, resulted in `ERROR:only 14000 out of 39583 10x biotype genes, results in 0.353687188944749 (0.40 threshold)`
- Ran notebook using AnnData with < threshold genes for human, resulted in `WARNING:17000 out of 39583 10x biotype genes, results in 0.42947730086148095 (0.60 threshold)`
- Ran notebook using AnnData with CSC matrix, resulted in `ERROR:X should be converted to csr sparse`
- Used AnnData with mixed organisms (mouse and human), resulted in: `ERROR:Multiple organisms found in obs metadata: ['NCBITaxon:9606', 'NCBITaxon:10090']`

### DIFF
--- a/cellxgene_resources/cellxgene_mods.py
+++ b/cellxgene_resources/cellxgene_mods.py
@@ -10,9 +10,11 @@ import scanpy as sc
 import squidpy as sq
 import subprocess
 import sys
+import requests
 from dataclasses import dataclass
 from pathlib import Path
 from scipy import sparse
+import cellxgene_schema.gencode as gencode
 
 
 portal_uns_fields = [
@@ -196,22 +198,29 @@ def determine_sparsity(x):
 def evaluate_sparsity(adata):
     max_sparsity = 0.5
 
+    valid = True
     sparsity = determine_sparsity(adata.X)
     report(f'X sparsity: {sparsity}')
     if sparsity > max_sparsity and type(adata.X) != sparse.csr_matrix:
-        report('X should be converted to sparse', 'WARNING')
+        report('X should be converted to sparse', 'ERROR')
+        valid = False
     
     if adata.raw:
         sparsity = determine_sparsity(adata.raw.X)
         report(f'raw.X sparsity: {sparsity}')
         if sparsity > max_sparsity and type(adata.raw.X) != sparse.csr_matrix:
-            report('raw.X should be converted to sparse', 'WARNING')
+            report('raw.X should be converted to sparse', 'ERROR')
+            valid = False
     
     for l in adata.layers:
         sparsity = determine_sparsity(adata.layers[l])
         report(f'layers[{l}] sparsity: {sparsity}')
         if sparsity > max_sparsity and type(adata.layers[l]) != sparse.csr_matrix:
-            report(f'layers[{l}] should be converted to sparse', 'WARNING')
+            report(f'layers[{l}] should be converted to sparse', 'ERROR')
+            valid = False
+
+    if valid:
+        report('all matrices are csr', 'GOOD')
 
 
 def evaluate_data(adata):
@@ -888,3 +897,61 @@ def evaluate_donors_sex(adata):
           )
 
         return donor_sex_df, dp
+
+
+def evaluate_var_df(adata):
+    """
+    Use single-cell-curation classes and fuctions and report warning/error for organism specific minimum number of gene features. Also, this function 
+    will look that var contains features from only a single organism.
+
+    :param obj adata: AnnData that is being curated
+
+    :return logging: Raises WARNING or ERROR if the number of genes in adata.var and adata.raw.var are fewer than the threshold of 40% or 60%, respectively,
+    of 10x preselected biotype of genes. Will also raise ERROR if dataset contains more than a singel organism.
+    """
+    accepted_biotypes = [
+        'protein_coding',
+        'protein_coding_LoF',
+        'lncRNA',
+        'IG_C_gene',
+        'IG_D_gene',
+        'IG_J_gene',
+        'IG_LV_gene',
+        'IG_V_gene',
+        'IG_V_pseudogene',
+        'IG_J_pseudogene',
+        'IG_C_pseudogene',
+        'TR_C_gene',
+        'TR_D_gene',
+        'TR_J_gene',
+        'TR_V_gene',
+        'TR_V_pseudogene',
+        'TR_J_pseudogene'
+    ]
+
+
+    # Check that this is single organism both in metadata and var index, exit function as does not make sense to check genes with multiple organisms
+    obs_organisms = adata.obs['organism_ontology_term_id'].unique()
+    var_organisms = {gencode.get_organism_from_feature_id(id).value for id in adata.var.index.to_list()}
+    if len(obs_organisms) > 1:
+        report(f'Multiple organisms found in obs metadata: {obs_organisms}', 'ERROR')
+        return
+    elif len(var_organisms) > 1:
+        report(f'Multiple organisms found in var index: {obs_organisms}', 'ERROR')
+        return
+    else:
+        report(f'Single organism found: {var_organisms}', 'GOOD')
+
+    # Check the number of genes threshold base on biotype per specific organism
+    organism = adata.obs['organism_ontology_term_id'].unique()[0]
+    org_obj = [i for i in gencode.SupportedOrganisms if i.value==organism][0]
+    gene_checker = gencode.GeneChecker(org_obj)
+    num_genes_biotype = len([i for i in gene_checker.gene_dict.keys() if gene_checker.gene_dict[i][2] in accepted_biotypes])
+
+    fraction = len(adata.var.index)/num_genes_biotype 
+    if fraction < 0.4:
+        report(f'only {len(adata.var.index)} out of {num_genes_biotype} 10x biotype genes, results in {fraction} (0.40 threshold)', 'ERROR')
+    elif fraction < 0.6:
+        report(f'{len(adata.var.index)} out of {num_genes_biotype} 10x biotype genes, results in {fraction} (0.60 threshold)','WARNING')
+    else:
+        report(f'{len(adata.var.index)} genes out of possible {num_genes_biotype} present', 'GOOD')
--- a/cellxgene_resources/cellxgene_mods.py
+++ b/cellxgene_resources/cellxgene_mods.py
@@ -220,7 +220,7 @@ def evaluate_sparsity(adata):
             valid = False
 
     if valid:
-        report('all matrices are csr', 'GOOD')
+        report('all matrices have passed checks', 'GOOD')
 
 
 def evaluate_data(adata):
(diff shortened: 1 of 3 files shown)


cellxgene_resources/cellxgene_mods.py
```diff
@@ -888,3 +897,61 @@ def evaluate_donors_sex(adata):
...
+    gene_checker = gencode.GeneChecker(org_obj)
+    num_genes_biotype = len([i for i in gene_checker.gene_dict.keys() if gene_checker.gene_dict[i][2] in accepted_biotypes])
+
+    fraction = len(adata.var.index)/num_genes_biotype 
+    if fraction < 0.4:
+        report(f'only {len(adata.var.index)} out of {num_genes_biotype} 10x biotype genes, results in {fraction} (0.40 threshold)', 'ERROR')
+    elif fraction < 0.6:
+        report(f'{len(adata.var.index)} out of {num_genes_biotype} 10x biotype genes, results in {fraction} (0.60 threshold)','WARNING')
+    else:
+        report(f'{len(adata.var.index)} genes out of possible {num_genes_biotype} present', 'GOOD')
```
jahilton, line 957: I don't know if "possible" is the correct term here (Would be odd to see "60k genes out of possible 33k present")
Maybe "x genes present, compared against y genes"?
Maybe also display the fraction as that's what is being weighed?

cellxgene_resources/cellxgene_mods.py
```diff
@@ -888,3 +897,61 @@ def evaluate_donors_sex(adata):
...
+        'TR_D_gene',
+        'TR_J_gene',
+        'TR_V_gene',
+        'TR_V_pseudogene',
+        'TR_J_pseudogene'
+    ]
+
+
+    # Check that this is single organism both in metadata and var index, exit function as does not make sense to check genes with multiple organisms
+    obs_organisms = adata.obs['organism_ontology_term_id'].unique()
```
jahilton, line 934: cleaner reporting if this is `.tolist()`

cellxgene_resources/cellxgene_mods.py
```diff
@@ -888,3 +897,61 @@ def evaluate_donors_sex(adata):
...
+    var_organisms = {gencode.get_organism_from_feature_id(id).value for id in adata.var.index.to_list()}
+    if len(obs_organisms) > 1:
+        report(f'Multiple organisms found in obs metadata: {obs_organisms}', 'ERROR')
+        return
+    elif len(var_organisms) > 1:
+        report(f'Multiple organisms found in var index: {obs_organisms}', 'ERROR')
```
jahilton, line 940: prints obs_organisms but should be var_organisms

also `NCBITaxon:2697049` (covid) should be OK as an additional organism in var

jahilton (changes requested):

cellxgene_resources/cellxgene_mods.py
jychien: How about this:
`report(f'{len(adata.var.index)} genes present, compared against {num_genes_biotype} 10x biotype genes; fraction: {fraction}', 'GOOD')`

cellxgene_resources/cellxgene_mods.py
jychien: Thanks for catching this.

cellxgene_resources/cellxgene_mods.py
```diff
@@ -196,22 +198,29 @@ def determine_sparsity(x):
...
     for l in adata.layers:
         sparsity = determine_sparsity(adata.layers[l])
         report(f'layers[{l}] sparsity: {sparsity}')
         if sparsity > max_sparsity and type(adata.layers[l]) != sparse.csr_matrix:
-            report(f'layers[{l}] should be converted to sparse', 'WARNING')
+            report(f'layers[{l}] should be converted to csr sparse', 'ERROR')
+            valid = False
+
+    if valid:
+        report('all matrices are csr', 'GOOD')
```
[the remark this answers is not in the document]
jychien: Yes! Oops, I'll update the wording on the report. I'll have it just print an "all clear", but let me know if you would like it to evaluate what types of matrices are in the h5ad.

cellxgene_resources/cellxgene_mods.py
```diff
@@ -10,9 +10,11 @@
 import squidpy as sq
 import subprocess
 import sys
+import requests
```
brian-mott, line 13: is requests used anywhere? looks like it was inadvertently added

cellxgene_resources/cellxgene_mods.py
```diff
@@ -888,3 +897,63 @@ def evaluate_donors_sex(adata):
...
+        'TR_J_gene',
+        'TR_V_gene',
+        'TR_V_pseudogene',
+        'TR_J_pseudogene'
+    ]
+
+
+    # Check that this is single organism both in metadata and var index, exit function as does not make sense to check genes with multiple organisms
+    obs_organisms = adata.obs['organism_ontology_term_id'].unique().tolist()
+    var_organisms = {gencode.get_organism_from_feature_id(id).value for id in adata.var.index.to_list()}
```
brian-mott, line 935: `get_organism_from_feature_id()` returns a `SupportedOrganism` enum or `None`, so if `evaluate_var_df()` is called when `adata.var` is indexed on gene symbol and/or deprecated Ensemble IDs have not been filtered, then this will raise an `AttributeError`

There could be logic to deal with this, try to catch this exception and print an error to index on ID and filter, or place the call to `evaluate_var_df` further in the curation notebook when var is expected to be indexed and filtered appropriately

cellxgene_resources/cellxgene_mods.py
```diff
@@ -888,3 +897,63 @@ def evaluate_donors_sex(adata):
...
+    if len(obs_organisms) > 1:
+        if 'NCBITaxon:2697049' not in obs_organisms:
+            report(f'Multiple organisms found in obs metadata: {obs_organisms}', 'ERROR')
+            return
+    elif len(var_organisms) > 1:
```
brian-mott, line 940: Might also be nice to rearrange this section to report both obs and var organism errors then return early. When there's multiple organisms in obs and var, then only the error report for obs will show; `evaluate_var_df` will have to be called again with obs fixed to get the var organism error

jychien:
@jahilton @brian-mott @Jchaffer787 Looks like I've missed many edge cases. Thank you for all the comments. I'll take a deep dive today and some out with something more robust tomorrow. If you find any more issues, please add to the comments.

Jchaffer787:
Nothing to add regarding code, looks like I was a little late to the party.
I was able to run the private fly dataset through as a test and output looks good:
```
GOOD:Single organism found: {'NCBITaxon:7227'}
GOOD:12553 genes present, compared against 13986 10x biotype genes; fraction: 0.8975403975403975
```
Could maybe be a space after 'GOOD:', but that's certainly a nitpick.

Jchaffer787 (commented):
Nothing to add regarding code, looks like I was a little late to the party.
I was able to run the private fly dataset through as a test and output looks good:
```
GOOD:Single organism found: {'NCBITaxon:7227'}
GOOD:12553 genes present, compared against 13986 10x biotype genes; fraction: 0.8975403975403975
```
Could maybe be a space after 'GOOD:', but that's certainly a nitpick.

cellxgene_resources/cellxgene_mods.py
jychien: Good point. I've updated logic, and will check both obs and var before breaking out of evaluate_var_df if there is an error.

cellxgene_resources/cellxgene_mods.py
jychien: Good point. For human, there are some genes that have Ensembl ID for both gene name and gene ID. So, I will check that None does not exist in var_organisms. Let me know if there is some edge case that I am not catching.

jychien: Sounds good! Will return early and exit out of evaluate_var_df() contains gene symbols or any Ensembl IDs that are not on the approved gene lists.

cellxgene_resources/cellxgene_mods.py
```diff
@@ -888,3 +897,63 @@ def evaluate_donors_sex(adata):
...
+        if 'NCBITaxon:2697049' not in obs_organisms:
+            report(f'Multiple organisms found in var index: {var_organisms}', 'ERROR')
+            return
+    else:
+        report(f'Single organism found: {var_organisms}', 'GOOD')
```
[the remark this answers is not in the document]
jychien: Will need to incorporate logic for Covid. If var is covid+human and obs is just human, then still need to catch this error.

jychien:
Alrighty. Here are things updated:

- Dropped import requests. Thanks for catching, I initially was going to go through gencode files in the single cell repo.
- Fixed Visium_assays embeddings and image check
- Added space after "GOOD:|ERROR:|WARNING:"
- Multiple organisms not results on check for both var and obs:
> ERROR: Multiple organisms found in obs metadata: ['NCBITaxon:9606', 'NCBITaxon:10090']
ERROR: Multiple organisms found in var index: ['NCBITaxon:10090', 'NCBITaxon:9606']

- If non approved gene in var index, evaluate_var_df() will report and skip any other checks:
> ERROR: Features in var.index are gene symbols and/or contain deprecated Ensembl IDs

- Allow for covid genes in var and still have it be considered single organism:
> There are covid genes present in var
GOOD: Single organism found: ['NCBITaxon:9606']

- Will ERROR is organism between var and obs are different:
> ERROR: Different organisms found between var index and obs metadata: NCBITaxon:9606, NCBITaxon:10090

- Will ERROR Is there is covid in obs metadata:
> ERROR: Covid is found in obs metadata

jahilton (approved):
lgtm

brian-mott (approved):
Looks good to me too, thanks Jenny for all the hard work!

Jchaffer787:
LGTM! Good output from fly file:
```
GOOD: Single organism found: ['NCBITaxon:10090']
GOOD: 22354 genes present, compared against 33989 10x biotype genes; fraction: 0.6576833681485187
```
Great work Jenny!!

Jchaffer787 (approved):

PR #238, opened 2025-03-26
upload to cxg via manifest

jahilton:
This doesn't include fragments file upload, which requires an additional entry in the manifest. Just might get messy when trying to account for fresh upload of new h5ad + fragments vs adding fragments to h5ad already uploaded
So will consider options for including those & submit a separate PR, but much less urgent


jychien:
Looks good to me!

jychien (approved):

brian-mott (approved):
Looks like this works on dev, I don't think this is on prod yet. I was able to upload a file to a private dev collection; it errored further in the validation process but this seemed to be unrelated to the upload process.

PR #241, opened 2025-04-05
Tools 253 update sample notebook

jychien:
Things updated for **cellxgene_mods.py**:
- Created concatenated genes_approved.csv.gz file of all organisms
- New code to generate genes_approved.csv.gz file in sample code notebook
- **map_filter_gene_ids()**: now looks for a gzipped file
- add check for cell type 'unknown' and will warn if larger than 20 (please let me know if this should be changed)
- **symbols_to_ids():**
  - will now expect to have genes_approved.csv.gz in ref_files directory, or else error out
  - new logic to find gene id based on symbol: 
    - split the symbol in genes_approved.csv.gz by '_'
    - try and match given symbol to the split gene symbol in genes_approved.csv.gz
    - will still try the lower case version of gene symbol, even though other organisms (such as drosophila) has mixed upper/lower case genes. This is because will still want to catch the mouse genes
  - for all the gene IDs found in genes_approved.csv.gz to have a matching gene symbol, it will check to see if any are found in var index
- `evaluate_10x_barcodes()` will print warning if no barcodes are found in obs index and then subsequently will not run `parse_barcode_df()` in curation notebook
  
 Updates for curation sample code: 
 - using cellxgene_schema.gencode package to get gencode metadata for all supported organisms

Have ran a drosophila, mouse, and human dataset through.

### DIFF
--- a/cellxgene_resources/cellxgene_mods.py
+++ b/cellxgene_resources/cellxgene_mods.py
@@ -387,12 +387,15 @@ def evaluate_obs_schema(obs, labels=False):
         for o in portal_obs_fields:
             if o in obs.keys():
                 report(f'schema conflict - {o} in obs\n', 'ERROR')
-    if 'cell_type_ontology_term_id' in obs.columns\
-            and 'unknown' in obs['cell_type_ontology_term_id'].unique()\
-            and len([i for i in obs['assay_ontology_term_id'].unique() if i in ['EFO:0022860','EFO:0022859','EFO:0022857']])==0:
-        num_unknown = obs[obs['cell_type_ontology_term_id']=='unknown'].shape[0]
-        if num_unknown> 20:
-            report(f'There {num_unknown} cells of unknown cell type in dataset.', 'WARNING')
+    if 'cell_type_ontology_term_id' in obs.columns and 'unknown' in obs['cell_type_ontology_term_id'].unique():
+        if 'in_tissue' in obs.columns:
+            num_unknown = obs.loc[(obs['in_tissue']==1) & (obs['cell_type_ontology_term_id']=='unknown')].shape[0]
+            perc_unknown = 100*(num_unknown/obs.loc[obs['in_tissue']==1].shape[0])
+        else:
+            num_unknown = obs[obs['cell_type_ontology_term_id']=='unknown'].shape[0]
+            perc_unknown = 100*(num_unknown/obs.shape[0])
+        if num_unknown > 20:
+            report(f'{num_unknown} ({perc_unknown}%) cells are cell_type:unknown.', 'WARNING')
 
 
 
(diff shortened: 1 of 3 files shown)


cellxgene_resources/cellxgene_mods.py
```diff
@@ -384,6 +387,13 @@ def evaluate_obs_schema(obs, labels=False):
         for o in portal_obs_fields:
             if o in obs.keys():
                 report(f'schema conflict - {o} in obs\n', 'ERROR')
+    if 'cell_type_ontology_term_id' in obs.columns\
+            and 'unknown' in obs['cell_type_ontology_term_id'].unique()\
+            and len([i for i in obs['assay_ontology_term_id'].unique() if i in ['EFO:0022860','EFO:0022859','EFO:0022857']])==0:
```
jahilton, line 392: instead of dropping Visium Datasets from the check, why not drop in_tissue:0 obs then do the same check?
could just check if in_tissue in obs.columns to avoid future update of EFO terms

cellxgene_resources/cellxgene_mods.py
```diff
@@ -384,6 +387,13 @@ def evaluate_obs_schema(obs, labels=False):
...
+        num_unknown = obs[obs['cell_type_ontology_term_id']=='unknown'].shape[0]
+        if num_unknown> 20:
+            report(f'There {num_unknown} cells of unknown cell type in dataset.', 'WARNING')
```
jahilton, line 395: Could be "There **are**...", but I think you could aim for more succinct - {num_unknown} cells are cell_type:unknown

jahilton: I would actually include % of unknowns in the report
So {num_unknown} ({perc_unknown}%) cells are cell_type:unknown

cellxgene_resources/cellxgene_mods.py
jychien: Good idea! Update for script to check visium using in_tissue rather than efo and subset to in_tissue:1 for cell_type unknown check

cellxgene_resources/cellxgene_mods.py
jychien: Thanks!  updated logging to be more concise and include percentage.

Jchaffer787 (approved):
> add check for cell type 'unknown' and will warn if larger than 20 (please let me know if this should be changed)

Tested, confirmed to work as expected. Output is `WARNING: 7415 (10.836049043534175%) cells are cell_type:unknown.`
Could maybe round the percentage output.

> symbols_to_ids(): will now expect to have genes_approved.csv.gz in ref_files directory, or else error out

Tested, confirmed to work as expected.

> evaluate_10x_barcodes() will print warning if no barcodes are found in obs index and then subsequently will not run 
parse_barcode_df() in curation notebook

Tested, confirmed to work as expected.

Otherwise, not seeing anything off in the code, LGTM!

PR #242, opened 2025-04-16
fix bug for looking at barcodes for barcode checker

jychien:
Things to note:
- I've updated `evaluate_10x_barcodes()` to look at obs['barcode'] for the warning. 
- I've also just moved the check for `barcode_df` for None into `parse_barcode_df()` so to avoid any complications of printing to screen in if-statement in the notebook

Tested:
- adata that with barcodes
- adata without barcodes: "WARNING: Did not find any barcodes in obs index, cannot evaluate barcodes"


brian-mott (approved):
Works on the initial dataset that caused the error and another dataset I just tried that is 10x 3' v3

Looks good, thanks for the quick fix!

PR #243, opened 2025-04-24
Fragments upload

jahilton:
Allow for ease of including optional fragments file with any Dataset upload
Allow for submitting fragments file with a local .h5ad or with existing .h5ad already in CxG


cellxgene_resources/upload_local.ipynb
```diff
@@ -96,45 +98,70 @@
...
-    "            print('Repeated dataset:' + d)"
+    "all_ids = [d['dataset_id'] for d in datasets]\n",
+    "for d in datasets:\n",
+    "    if d['anndata'] != 'existing' and not os.path.exists(d['anndata']):\n",
+    "        print('Invalid file:' + d['anndata'])\n",
+    "    if 'fragments' in d and not os.path.exists(d['fragments']):\n",
+    "        print('Invalid file:' + d['fragments'])\n",
+    "    if d['dataset_id'] != 'new':\n",
+    "        if d['dataset_id'] not in titles:\n",
+    "            print('Invalid dataset:' + d['dataset_id'])\n",
```
brian-mott, line 155: If the dataset_id does not yet exist, then there will be KeyErrors for checking the dataset_id in the `status` dictionary. I would put a continue after the this line and then move this whole if block (if d['dataset_id'] != 'new') to be the last large if block check.

That way all the other checks can occur, if the dataset id is not in `titles`, then the following checks that depend on dataset_id are skipped without erroring

cellxgene_resources/upload_local.ipynb
```diff
@@ -96,45 +98,70 @@
...
-    "for f,d in datasets.items():\n",
-    "    if not os.path.exists(f):\n",
-    "        print('Invalid file:' + f)\n",
-    "    if d != 'new':\n",
-    "        if d not in coll_datasets:\n",
-    "            print('Invalid dataset:' + d)\n",
-    "        if list(datasets.values()).count(d) > 1:\n",
-    "            print('Repeated dataset:' + d)"
+    "all_ids = [d['dataset_id'] for d in datasets]\n",
+    "for d in datasets:\n",
```
brian-mott, line 148: I think an enumerate would be helpful here, for the two errors that start with 'Must define' The other error messages give a file or dataset_id or something to check and correct; those two errors aren't specific enough to pinpoint which dictionary in the datasets list to correct. So with enumerate, you could include that index in the message with something like:
`f"Must define fragments if revising an existing matrix for datasets[{index}]"`

cellxgene_resources/upload_local.ipynb
```diff
@@ -96,45 +98,70 @@
...
    "execution_count": null,
-   "id": "8be0c822",
+   "id": "159f2bde-5bcf-4430-967a-337ffe7e3bf7",
    "metadata": {},
    "outputs": [],
    "source": [
-    "datasets = {\n",
-    "    directory + 'my_matrix_revised.h5ad': 'new'\n",
-    "}"
+    "datasets = [\n",
```
brian-mott, line 114: This dictionary of lists seems fine to me, can see what others think. Also not complicated to do list/dictionary comprehensions to create a large `datasets` list of dicts, which I sometimes do if there are many files to update/upload at once.

cellxgene_resources/upload_local.ipynb
```diff
@@ -96,45 +98,70 @@
...
    "source": [
-    "for f,d in datasets.items():\n",
-    "    if not os.path.exists(f):\n",
-    "        print('Invalid file:' + f)\n",
-    "    if d != 'new':\n",
-    "        if d not in coll_datasets:\n",
-    "            print('Invalid dataset:' + d)\n",
-    "        if list(datasets.values()).count(d) > 1:\n",
-    "            print('Repeated dataset:' + d)"
+    "all_ids = [d['dataset_id'] for d in datasets]\n",
```
brian-mott, line 147: And just a nit for this section in general: it's a little easier to read the error print outs with a space after the colon in the first phrase of the string.

cellxgene_resources/upload_local.ipynb
jahilton: I also moved the 'repeated datasets' check up so that will always occur

cellxgene_resources/upload_local.ipynb
jahilton: nits are welcome and appreciated

PR #244, opened 2025-05-20
Tools 261 move organism to uns

jychien:
Ran standard test datasets in test_processed_matrix_files.txt and all moved organism_ontology_term_id to adata.uns.


Jchaffer787:
Code looks good, and standard tests all passed. Checked organism_ontology_term_id does move to adata.uns.
LGTM!

PR #245, opened 2025-05-23
Tools 259 allow disease lists

Jchaffer787:
Change disease handling to allow for disease lists.

gather_pooled_metadata was modified to update unknown check to handle lists for both diseases.term_id and diseases.term_name. Also removed old multiple values check and warning.

clean_list was modified to allow disease_ontology_term_id to be a list if disease_found is a list.

Tested for non-pooled on mocked file on demo (LATDF006UTH) where experimental disease was set to a list of values that some of the donors also had contained in the list of reported diseases.

For pooled handling, modified all donors in a library in LATDF675NTY to have multiple diseases, and then set experimental disease to same list. Result was as expected, presence of || separated string containg both diseases term ids.

### DIFF
--- a/scripts/flattener.py
+++ b/scripts/flattener.py
@@ -149,7 +149,7 @@ def clean_list(lst, exp_disease, glob):
 	disease_found = [i for i in exp_disease_list if i in lst]
 	if disease_found:
 		if len(disease_found) > 1:
-			disease = ' || '.join(disease_found.split(','))
+			disease = ' || '.join(disease_found)
 		else:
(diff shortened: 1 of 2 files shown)


jychien:
On demo:
- Added MONDO:0004496 to all donor associated with library LATLB203CON from ProcMatrixFile LATDF675NTY. And added experimental disease LATDF675NTY. Looks good!
Screenshot 2025-06-02 at 3 55 52 PM

- Still need to see test erroring.

PR #246, opened 2025-05-23
TOOLS-260_change_ethnicity_list_delimiter

Jchaffer787:
Modified delimiter used to join ethnicity list, and corrected handling of list in gather_pooled_metadata.

Tested on modified version of LATDF006UTH for unpooled multiple ethnicities, and got expected outcome.

Unfortunately, the only viable file for testing pooled multiple ethnicities is LATDF675NTY, which also has multiple dieases. Which errors out on this branch.
We should test this file on the TOOLS-251 branch, since it will cover both.

Also ran through flattener testing script with all successes

### DIFF
--- a/scripts/flattener_mods/gather.py
+++ b/scripts/flattener_mods/gather.py
@@ -10,7 +10,7 @@ from cellxgene_ontology_guide.ontology_parser import OntologyParser
 from dataclasses import dataclass
 
 # add some constant lookup to keep in sync with current schema version
-ONTOLOGY_PARSER = OntologyParser(schema_version=f"v5.2.0")
+ONTOLOGY_PARSER = OntologyParser(schema_version=f"v5.3.0")
 
 # Backtracking to scripts folder to import lattice.py
 sys.path.insert(0, '../')
@@ -219,7 +219,7 @@ def gather_metdata(obj_type, properties, values_to_add, objs, connection):
 					else:
 						ethnicity_list.append(ethnicity_dict.get('term_id'))
 				ethnicity_list.sort()
-				value = ','.join(ethnicity_list)
+				value = ' || '.join(ethnicity_list)
 				latkey = (obj_type + '_' + prop).replace('.','_')
 				key = constants.PROP_MAP.get(latkey, latkey)
 				values_to_add[key] = value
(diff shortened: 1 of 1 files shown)


jychien:
Tested:
- /processed-matrix-files/LATDF216UIK/: 
  - updated pooled donor LATHD952BYS to have also HANCESTRO:0015:
    1. Getting: `Pooled ethnicities '{'HANCESTRO:0005', 'HANCESTRO:0005 || HANCESTRO:0015'}' for 'pooled [D9,D7,D8,D10,D6]', setting to 'unknown'`, which is the correct answer, but could be coincidence (@brian-mott), as it looks as if the following line is looking for a list: https://github.com/Lattice-Data/lattice-tools/blob/7028f435ab8caf25d139155f14cce627db66ac7d/scripts/flattener_mods/gather.py#L293
    2. So, I updated pooled donor LATHD952BYS to have also HANCESTRO:0462 (British, which is a child of European) and also reverted back to ',' separator, still getting `WARNING: Pooled ethnicities '{'HANCESTRO:0005', 'HANCESTRO:0005,HANCESTRO:0462'}' for 'pooled [D9,D7,D8,D10,D6]', setting to 'unknown'`
  - updated donor LATHD386HBY to have also HANCESTRO_0009:
    - D2 accurately has `HANCESTRO:0005 || HANCESTRO:0009` 

So, looks like find_common_ontology_term does account for pooled samples that are ['HANCESTRO:0005,HANCESTRO:0462'], but not for pooled samples that are ['HANCESTRO:0005,HANCESTRO:0462','HANCESTRO:0005']. That does seem okay to me. My thinking is that if, for example, a donor is listed as 2 ethnicities, it is most likely that they do not have a common ancestor. For example, it would never actually be the case **b** that I had listed above, where a donor is European and British. So, it seems safe to assume that a donor of multiple ethnicities, if pooled into a sample with other donors, would result in 'unknown' (unless the other donors are the exact same ethnicities). 

@Jchaffer787 @brian-mott Thoughts? Sorry it's so confusing. I tried to be as concise as possible. The other option is to add the swap from `,` to ` || ` to `clean_obs()` (flattener #L479), and then add the logic to find_common_ontology_term to go through sublists to find common ancestors. And that seems overkill.

jychien (approved):

PR #247, opened 2025-06-04
Tools 251 update flattern cxg 6 0

jychien:
Log information about testing here. Only merge when schema 6.0 migration has started.

### DIFF
--- a/scripts/flattener.py
+++ b/scripts/flattener.py
@@ -225,7 +225,7 @@ def report_diseases(mxr_df, exp_disease):
 		mxr_df['reported_diseases'].replace({'[]':'none'}, inplace=True)
 	mxr_df.loc[mxr_df['pooled'],'reported_diseases'] = 'pooled ' + mxr_df['reported_diseases']
 
-	# calculate disease_ontology_term_id frorm sample and donor disease, using list manipulations to start with the unique disease found in both
+	# calculate disease_ontology_term_id from sample and donor disease, using list manipulations to start with the unique disease found in both
 	if exp_disease == fm.UNREPORTED_VALUE:
 		mxr_df['disease_ontology_term_id'] = ['PATO:0000461'] * len(mxr_df.index)
 	else:
--- a/scripts/flattener_mods/gather.py
+++ b/scripts/flattener_mods/gather.py
@@ -386,7 +386,6 @@ def gather_pooled_metadata(obj_type, properties, values_to_add, objs, connection
 					if len(v) == 1 and v[0] == 'unknown':
 						values_df.loc[key,ident] = 'unknown'
 					elif 'unknown' in v:
-						#Actually should error here? Cause then single donor has mix of unknown and disease
 						values_df.loc[key,ident] = 'unknown'
 					else:
 						value = ' || '.join(v)
@@ -409,7 +408,6 @@ def gather_pooled_metadata(obj_type, properties, values_to_add, objs, connection
 					if len(v) == 1 and v[0] == 'unknown':
 						values_df.loc[key,ident] = 'unknown'
 					elif 'unknown' in v:
-						#Actually should error here? Cause then single donor has mix of unknown and disease
 						values_df.loc[key,ident] = 'unknown'
 					else:
 						value = ' || '.join(v)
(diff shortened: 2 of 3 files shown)


scripts/flattener.py
```diff
@@ -176,26 +194,55 @@ def concat_list(anndata_list, column, uns_merge):
...
+	mxr_df['reported_diseases'] = mxr_df['reported_diseases'].astype('str')
 	total_reported = mxr_df['reported_diseases'].unique()
 	if len(total_reported) == 1:
 		if total_reported[0] == '[]':
 			mxr_df['reported_diseases'] = '[]'
 	elif '[]' in total_reported:
 		mxr_df['reported_diseases'].replace({'[]':'none'}, inplace=True)
+	mxr_df.loc[mxr_df['pooled'],'reported_diseases'] = 'pooled ' + mxr_df['reported_diseases']
 
+	# calculate disease_ontology_term_id frorm sample and donor disease, using list manipulations to start with the unique disease found in both
```
Jchaffer787, line 228: Frorm -> from :)

scripts/flattener_mods/gather.py
```diff
@@ -375,6 +375,52 @@ def gather_pooled_metadata(obj_type, properties, values_to_add, objs, connection
...
+			latkey = (obj_type + '_' + prop).replace('.','_')
+			key = constants.PROP_MAP.get(latkey, latkey)
+			for obj in objs:
+				ident = obj.get('@id')
+				v = get_value(obj, prop)
+				if isinstance(v, list):
+					if len(v) == 1 and v[0] == 'unknown':
+						values_df.loc[key,ident] = 'unknown'
+					elif 'unknown' in v:
+						#Actually should error here? Cause then single donor has mix of unknown and disease
```
Jchaffer787, line 389: Could be an edge case where disease for one of the donors in a pooled sample is ['unknown','disease1'] and this would just set it to 'unknown' for that donor. Is this how we would want to handle that case? It'd be strange to assign both a disease and unknown to the same donor, but we've seen stranger.

scripts/flattener_mods/gather.py
```diff
@@ -375,6 +375,52 @@ def gather_pooled_metadata(obj_type, properties, values_to_add, objs, connection
...
```
Jchaffer787, line 412: Same as comment above.

Jchaffer787:
Did some testing with non-pooled disease on LATDF006UTH by changing up experimental_disease list in processed matrix file. Got expected outputs each time:
`disease_ontology_term_id
MONDO:0005575 || MONDO:0024880                     61404
PATO:0000461                                       30099
MONDO:0005575                                      23897
MONDO:0006698                                      22130
MONDO:0015074 || MONDO:0006698 || MONDO:0005148    14539
MONDO:0005148                                      13927
MONDO:0006698 || MONDO:0005148                      1602`

Also re-ran pooled ethnicity testing on LATDF675NTY, getting expected output.

Did a quick modification of LATDF675NTY to check pooled multiple disease, and also got expected output.

Tomorrow I'm going to continue testing with putting 'unknown' in for disease to see how that is handled. Making a note here that the following donors have been modified along with the processed matrix file:
/human-postnatal-donors/LATHD091YJE/
/human-postnatal-donors/LATHD141DAL/
/human-postnatal-donors/LATHD547TLR/
/human-postnatal-donors/LATHD688SLQ/
/human-postnatal-donors/LATHD711CYP/

Jchaffer787:
Tested putting empty diseases field in for a single donor in a pooled sample, got expected error:
`ERROR: experimental diseasee MONDO:0004496 inconsistent across donors/samples pooled [K107238,K108019,K107957,K107889,K108056]`

Putting in empty diseases field for all donors in a pooled sample resulted in expected PATO term for normal assignment.

@jychien Other than removing the left over comments about the error, all looks good!

jychien:
@Jchaffer787 Thank you much for the thorough testing! I've pushed the changes that you suggested for logging and also schema version, as I think it must have been dropped somewhere along the way.

We will still need to update gather.py ONTOLOGY_PARSER, but if we still want to use and test this flattener, will need to leave it at 5.3.0 for the time being.

jychien:
Updated ONTOLOGY_PARSER to 6.0.0.

Ran and flattened LATDF329OGL and passed 6.0.0 validation.

PR #249, opened 2025-06-19
Change to values of make_numeric

Jchaffer787:
'suspension_percent_cell_viability' has changed to 'cell_viability_percentage', make_numeric has been changed to reflect this, so that 'cell_viability_percentage' is still numeric.


Jchaffer787:
Tested using LATDF817DNZ as Jenny recommended, and got dtype of float64

@jychien Looks like this is good to go

jychien:
Looks good! Thanks, Jim!

PR #250, opened 2025-06-20
Tools 264 fix qa notebook for descendant organisms

Jchaffer787:
Adding handling of organisms with descendants to evaluate_var_df()

Will now check if uns organism is a descendant of var organism if uns organism is in the has_descendants list.

Have tested on the pig dataset from the jira ticket.

Note that I've added Corinns changes for the uns organism so the merge for that will be more straightforward, so it's checking for organism in uns.

### DIFF
--- a/cellxgene_resources/cellxgene_mods.py
+++ b/cellxgene_resources/cellxgene_mods.py
@@ -21,11 +21,13 @@ import cellxgene_schema.schema as schema
 portal_uns_fields = [
     'citation',
     'schema_reference',
-    'schema_version'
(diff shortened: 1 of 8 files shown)


jychien:
Perfect! Thank you for adding ONTOLOGY_PARSER object. And the swap for line 994 makes sense to me!

I modified a mouse dataset with adata.uns['organism_ontology_term_id'] = 'NCBITaxon:947985'. The curation notebook ran through smoothly as expected. Thanks!

jychien (approved):

PR #254, opened 2025-07-16
Tools 267 3pv1 barcodes & feature_is_filtered check

corismall:


jychien:
Looks good!
- Ran a test dataset through with barcodes added to match 10x 3' v1 and it looked good! 
- dataset with >50% feature_is_filtered True came out with warning.

PR #264, opened 2025-09-19
Jason bcp

jychien:
@jahilton  Ready to merge to main to run first full set of Marson curated matrices


jahilton (changes requested):
Looks OK to me, but I would not submit the curated_matrix & expected_files ntbks as part of this PR. The curated_matrix ntbk should now be relegated to a testing framework used to develop the .py files, and the expected_files is unrelated to the curated matrix generation.

jychien:
@jahilton Good point. I've deleted the notebooks.

Jchaffer787:
@jychien Just pushed fixes for those issues as well as better handling for self_reported_ethnicity mapping.

jychien:
Nice, thanks @jahilton.
@Jchaffer787 Did you see any other issues?

Jchaffer787 (approved):

jychien:
@Jchaffer787 For dropping "n/a", I am trying to shift the use of "n/a" to "na". I know for your recent commit, though, it's "n/a" and it's just passed along and dropped anyways. Thoughts? Or, we can just merge this and iron this out the next round of work.

Jchaffer787:
@jychien So then would we not be expecting to see 'n/a' in the sheet at all? Or do we want to catch 'n/a' and convert to 'na'?

jychien:
@Jchaffer787 After thinking about it some more, let's move away from using "n/a". That way, we don't have to deal with any inconsistencies in the code. CxG requires "na", so let's stick with using "na" for all spreadsheet column.

PR #271, opened 2025-09-30
Concatenator Initial Merge

brian-mott:
See https://lattice.atlassian.net/browse/TOOLS-237 for more detailed information.

Merge initial set of tools for wrangling fragment files out of Lattice towards CELLxGENE. 

Entry point script in scripts/concatenator.py takes command line arguments in the style of the flattener (mode/env of prod or demo and processed matrix accession) to create a final filtered and concatenated fragment file associated with the given processed matrix. This is done in parallel as much as possible. The most efficient mode makes use of concurrent tsv file streaming (via a binary initially written in Go) to read and write tsv files at the same time, followed by parallel gzip compression with pigz. The Go binaries are included in this PR; pigz needs to be installed on the given machine to make use of parallel compression. Otherwise, the script will default to python multi-processing and gzip subprocesses for compression.

The design of the fragment processing step allows for selection of different workers; by default, this will be filtering against the obs index of the processed matrix file. A de-duplication worker can also be selected if validation reveals duplicate fragment file rows, or the Go binary processing can be selected as well. Further worker types can be added to handle other issues that might arise, such as lift over to another primary assembly, further investigation of validation errors, etc.

Extensive testing has been done on a number of processed matrix files, with further details found in TOOLS-237. Jennifer and Erica have both used this tool at least once or twice and provided very helpful feedback.


brian-mott:
Renamed to fragment_curator.py and places the 4 associated files in ./fragment_curator_mods. Reran with both python and golang binary filtering and everything seemed to work. I think I also updated all the correct documentation and comments to refer to the new name and to create the appropriate log file name.

PR #300, opened 2026-02-05
Update concurrent validation

brian-mott:
See TOOLS-282 for more details.

Updating `concurrent_validation.py` to be able to take CXG validation pre-analysis arguments, and to log validation output to file. Both these features should be very helpful for validating the numerous BCP curated matrices.

Tested locally and on JupyterHub with various h5ads within BCP and from other sources (CXG curation, flattener, etc).


brian-mott:
Discovered a bug where logging queue becomes overwhelmed with too many errors (over 32767) and drops validation errors. Now will show summary of validation in real time as subprocesses complete, will print only first 100 errors to terminal, and will log full validation errors to file

PR #323, opened 2026-06-10
Fix wafer summary report generation

gabdank:


gabdank:
@jychien I have fixed the erroneous reports produced for Psomagen orders with multiple wafers, that cause my miscommunication with Zohar. If you can try the new code to make sure it is working fine for you, I would merge.

jychien (approved):
@gabdank looks good to me! I ran multiple orders for multiple assays through from both CRO, and everything ran as expected. Thanks

PR #325, opened 2026-06-16
FASTQs and H5s data and metadata extraction from S3

gabdank:


jychien (approved):
@gabdank Worked well, thank you! I used it for my latest submission and it was helpful. I still had to do some additional random things, which are manageable, including:
- created alias from concatenation of library and splitting the url
- changed "CRISPR Guide Capture" to "guide count", and "Gene Expression" to "gene"
- make sure to import the tsv files, as crc64nvme_base64 could start with "="

PR #327, opened 2026-06-19
Tools 299 demultiplexing spreadsheet

Jchaffer787:
Change source of donor information for sample demultiplexing sheet from library donors to rawmatrixfile sample donors.
Remove logic checking for presence of samples field in rawmatrixfiles, as it is now a mandatory field.

### DIFF
--- a/TOOLS-285-DB2Flattener/DB2Flattener.py
+++ b/TOOLS-285-DB2Flattener/DB2Flattener.py
@@ -14,9 +14,6 @@ import DB2lattice
 class DB2Flattener:
     def __init__(self):
         # Setup connection
-        os.environ['DEMO_KEY'] = ''
-        os.environ['DEMO_SECRET'] = ''
-        os.environ['DEMO_SERVER'] = 'https://lattice-api-dev.demo.lattice-data.org'
         self.connection = DB2lattice.Connection('demo')
         
         # Initialize gatherer
@@ -67,7 +64,7 @@ class DB2Flattener:
             sample_df = sample_df.drop(columns=sample_cols_to_drop)
             
             print(f"Saving sample DataFrame to {sample_output}...")
-            sample_df.to_csv(sample_output, index=True)  # Keep index since it's the sample alias
+            sample_df.to_csv(sample_output, index=False)
             
             print(f"✅ Sample CSV file created: {sample_output}")
             print(f"   Rows: {len(sample_df)}")
@@ -106,7 +103,7 @@ class DB2Flattener:
     
     
     def create_dataframe(self, complete_data):
-        """Create main library/raw-file DataFrame and optional sample-indexed DataFrame"""
+        """Create main library/raw-file DataFrame and sample DataFrame keyed by raw matrix file"""
         libraries_data = complete_data['libraries']
         resolved_controlled_terms = complete_data['resolved_objects'].get('ControlledTerm', {})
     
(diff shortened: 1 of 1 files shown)


TOOLS-285-DB2Flattener/DB2Flattener.py
brian-mott: Just starting to review but these lines look like they are redundant and cause issues with defined env variables. They might be temporarily overwriting my env variable values during a flattener run and/or interfering with how the Connection object pulls in the env variables. I get this error: "ERROR: DEMO_KEY, DEMO_SECRET, DEMO_SERVER not all defined. Try 'conda env config vars list' to list existing variables"

If I delete these lines, then the flattener works.

Probably a convenient way to easily change these as you work across different demo instances, but I think these should just be deleted to not accidentally commit the key and secret to the repo.

TOOLS-285-DB2Flattener/DB2Flattener.py
```diff
@@ -121,235 +121,157 @@ def create_dataframe(self, complete_data):
...
-    
+                # Experimental condition information
+                'experimental_condition': self._join_unique([e.get('condition') for e in experimental_conditions]),
+                'experimental_condition_details': self._join_unique([e.get('text_value') for e in experimental_conditions]),
+        
+                # Genetic modification information
+                'genetic_modifications' : self._join_unique([g.get('strategy') for g in genetic_modifications]),
+        
+                # Cell type information from samples
+                'enriched_cell_types': self._get_cell_types_from_samples(samples,
```
brian-mott, line 251: This is a style thing and something we can figure out with linter rules, but I like formatting to align one indent level in, like:

                'enriched_cell_types': self._get_cell_types_from_samples(
                    samples, 
                    'enriched_cell_types', 
                    resolved_controlled_terms
                ),
`
instead of very far to the right indenting or lots of arguments on one line.
We can worry about this later

brian-mott (commented):
This looks good, left a few comments about the env variables and formatting, but looks like the logic works from what I can tell. @jychien can speak better to if the final output matches what is expected; I do at least see just one sex now for each donor.

jychien:
@Jchaffer787 The demultiplexing sex looks good! I am thinking a little more about what users will need in this multiplexing file:
- Example GSE that will eventually hold the demultiplexing file: https://www.ncbi.nlm.nih.gov/geo/query/acc.cgi?acc=GSE314342
- Example GSM that will hold the pooled GEO sample metadata: https://www.ncbi.nlm.nih.gov/geo/query/acc.cgi?acc=GSM9393828

I think it will be more usable if users are able to map metadata to the h5 file level (which would be a merge of the current two output csv files:
- each row will be a single h5 file
- each h5 will be mapped to the demultiplexed sample metadata

And then, we will do a groupby() on groupID for the GEO sample metadata.csv. I can get groupby() in for TOOLS-301.

TOOLS-285-DB2Flattener/DB2Flattener.py
Jchaffer787: Agreed, this was useful for when demo was changing a lot, but now not so much. I have removed these lines.

Jchaffer787:
I've made it so that metadata is mapped to the h5 level as Jenny described.
So now the _samples.csv contains all the metadata from the main dataframe, but also the demultiplexed sample metadata merged in.

jychien (approved):
@Jchaffer787 Looks good to me! Thanks!

PR #329, opened 2026-06-23
Tools294 qa ntbk v710

jahilton:
Updating various QA steps to account for new required & optional fields introduced in CxG schema 7.1

**Not included: evaluation of uns.genetic_perturbations, if present

### DIFF
--- a/cellxgene_resources/cellxgene_mods.py
+++ b/cellxgene_resources/cellxgene_mods.py
@@ -23,13 +23,12 @@ OBS_ONTOLOGY_LABELS_REQUIRED = [
     'assay', 'cell_type', 'development_stage', 'disease',
     'self_reported_ethnicity', 'sex', 'tissue'
 ]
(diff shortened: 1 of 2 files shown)


brian-mott (approved):
Looks good to me. I tried non-valid datasets at the start of curation, a valid dataset without labels, a dataset downloaded from CXG with labels, and one of the genetic perturbation fixtures with labels.

There are currently no genetic perturbation datasets on CXG, so the fixture with labels added locally is the closest I could get to mimic this use case. This lacks `adata.uns['citation']` but otherwise should have all the other portal-added fields. I think iterating through the expected uns keys is fine, instead of checking for presence first; I think we are always using `revise_cxg()` with datasets from the CXG and not with datasets we just locally labeled ourselves.

There might be other things that pop up as we go through curation but I think better to get us all using this than waiting for this merge.

PR #330, opened 2026-06-24
TOOLS-310 Restore sublibrary-level trimmer histograms 

gabdank:


jychien (approved):
@gabdank Looks good, thanks! Ran a scale and psomagen order.

PR #331, opened 2026-06-25
Bcp 145 add q30 stats

jychien:
@gabdank  Added logging and histogram for q30 values in Ultima raw log files:
- gather "PCT_PF_Q30_METRIC" metric from csv file
- optimized by multi threading reading from s3 files
- set cutoff as 65 so as to flag sublibraries with "red flag"
- only plot graph for plate when applicable and do not log all values


gabdank:
@jychien you need to rebase the branch onto main, before merging. Additionally, I am wondering if we need s3fs or can get away with boto3?

jychien:
@gabdank 
- thanks, I've rebased
- I had wondered about s3fs. It was introduced because I had found that the check was running slow with fsspec and wanted to multi-process the reading. This is what agent has said:

> **Why s3fs was chosen**
> From the BCP-145 work:
> - File-like streaming API (open() + line iteration)- > File-like streaming API (open() + line iteration)
> - One shared S3FileSystem for concurrent reads- > One shared S3FileSystem for concurrent reads
> - Region passed via client_kwargs, not by reusing the boto3 client (s3fs uses aiobotocore and rejects client=s3_client)- > Region passed via client_kwargs, not by reusing the boto3 client (s3fs uses aiobotocore and rejects client=s3_client)
>
> **Practical recommendation**
> - Keep s3fs if you’re happy with the current Q30 path and want to avoid rework; it’s a small, focused dependency (requirements.txt already has it).
> - Switch to boto3 if you want one fewer dependency and are fine updating _read_pct_q30_from_s3, tests (TestGatherPctQ30 mocks s3fs.S3FileSystem), and requirements.txt.
>
> Functionally, boto3 streaming is enough; s3fs is a convenience/optimization choice, not a hard requirement for the gatherer as a whole.

gabdank (approved):

PR #332, opened 2026-06-25
Tools312 flexv2 barcodes

jahilton:
add flex v2 and 3' v1 barcodes to master barcode list
update barcode parsing to account for flex


Jchaffer787 (approved):
code LGTM.
Ran some 3pv1 through to check, didn't have any flex v2 to check.

PR #333, opened 2026-06-29
Tools 301 add geo metadata columns

jychien:
@Jchaffer787 Here is the PR for the recent updates, with the main things being
- iterate through constants.py for metadata fields to extract
- extract libraries metadata from RawMatrixFile
- extract donor and sample metadata using demultiplexed sample at RawMatrixFile
- created df_utils.py to contain any code for dataframe manipulation logic
- create geo mapping to create geo csv from main_df

### DIFF
--- a/TOOLS-285-DB2Flattener/DB2Flattener.py
+++ b/TOOLS-285-DB2Flattener/DB2Flattener.py
@@ -211,7 +211,7 @@ class DB2Flattener:
                 if sample_obj:
                     alias = self._get_clean_alias(sample_obj)
                     sample_aliases.append(alias)
-            
+
             row = {
                 'raw_matrix_file_alias': self._get_clean_alias(raw_file),
                 'library_aliases': self._join_unique([self._get_clean_alias(lib) for lib in libraries]),
@@ -220,14 +220,16 @@ class DB2Flattener:
                 'library_protocols': self._join_unique([lib.get('protocol', '') for lib in libraries]),
                 'library_types': self._join_unique([lib.get('@type', [''])[0] if 
                                                     lib.get('@type') else '' for lib in libraries]),
+                'library_CRO_group_identifier': self._join_unique([lib.get('CRO_group_identifier', '') for lib in libraries]),
+                'library_construction_technology': self._join_unique([lib.get('library_construction_technology') for lib in libraries]),
                 
                 # Sample information
                 'raw_file_samples': self._join_unique(sample_aliases),
                 'library_sample_types': self._join_unique([s.get('@type', [''])[0] for s in samples]),
                 'selection_markers': self._join_unique([s.get('selection_markers', [''])[0] for s in samples]),
+                'sample_terms': self._join_unique([s.get('sample_terms') for s in samples]),
 
                 # Donor information
-                'donor_sexes': self._join_unique([d.get('sex', '') for d in donors]),
                 'donor_ethnicities': self._join_unique([
                         self._resolve_controlled_term(d.get('ethnicity'), resolved_controlled_terms)
                         for d in donors
--- a/TOOLS-285-DB2Flattener/DB2Flattener.py
+++ b/TOOLS-285-DB2Flattener/DB2Flattener.py
@@ -4,6 +4,7 @@ import sys
 import pandas as pd
 from datetime import datetime
 from DB2Gatherer import DB2Gatherer
+from constants import OBJECT_CONFIG
 
 # Add parent directory to path
 parent_dir = os.path.dirname(os.path.dirname(os.path.abspath(__file__)))
@@ -118,7 +119,7 @@ class DB2Flattener:
             if lib_data['library'].get('@id') in filtered_library_ids
         }
         
-        print(f"Creating DataFrame by {'raw matrix file'}...")
+        print("Creating DataFrame by raw matrix file...")
         
         rows = []
         sample_df = None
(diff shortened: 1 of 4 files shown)


TOOLS-285-DB2Flattener/df_utils.py
```diff
@@ -0,0 +1,151 @@
...
+    if not items:
+        return pd.NA
+    unique = list(dict.fromkeys(items))
+    return unique[0] if len(unique) == 1 else unique
+
+
+def collapse_dataframe(
+    df: pd.DataFrame,
+    group_col: str,
+    columns: list[str] | None = None,
```
Jchaffer787, line 49: I think its safe to drop this columns parameter.
It looks like columns parameter is for indicating all the columns in the dataframe that aren't the group column, but we can just make that calculation here in this method (as it's doing already). Doesn't make sense to calculate it every time we want to call this method and pass it as a parameter. Unless I'm missing something logically, or we need this for something in the future?

TOOLS-285-DB2Flattener/DB2Gatherer.py
Jchaffer787: This line can be removed, sample_refs is never used.

Jchaffer787: This line can also be removed, same reason as above, sample_refs is never used.

Jchaffer787: The all_samples parameter can be removed, not used.

Jchaffer787: This line, as well as controlled_terms, genetic_modifications, and experimental_conditions lines below can be deleted. Not used in flattener.py

TOOLS-285-DB2Flattener/DB2Flattener.py
```diff
@@ -4,12 +4,14 @@
 import pandas as pd
 from datetime import datetime
 from DB2Gatherer import DB2Gatherer
+from constants import OBJECT_CONFIG, PROP_MAP_GEO
+from df_utils import split_controlled_term_columns, collapse_dataframe
+import DB2lattice
 
 # Add parent directory to path
 parent_dir = os.path.dirname(os.path.dirname(os.path.abspath(__file__)))
 sys.path.append(parent_dir)
```
Jchaffer787: Can be dropped, no longer necessary since DB2lattice is now in same directory.

TOOLS-285-DB2Flattener/DB2Flattener.py
Jchaffer787: Can be deleted. Not used.

Jchaffer787: Can be deleted. Not used.

Jchaffer787: Can be deleted. Not used

TOOLS-285-DB2Flattener/DB2Flattener.py
```diff
@@ -329,27 +327,27 @@ def _get_diseases_from_samples(self, samples, controlled_terms):
...
         unique_items = sorted(set(filtered_items))
         return '; '.join(unique_items)
     
     def _join_list(self, items):
-        """Join list items directly, return 'n/a' if empty"""
+        """Join list items directly"""
         if not items:
-            return 'n/a'
+            return
         return '; '.join(items)
```
Jchaffer787: Cursor caught this one, but looks like we don't use _join_list anywhere, so can be dropped.

TOOLS-285-DB2Flattener/DB2Gatherer.py
Jchaffer787, line 129: resolve_referencs_for_samples needs to be adjusted so that it is no longer batch fetching Treatment, GeneticModification, or ExperimentalCondition.

Currently, this fetches all references, but we only want it to fetch HumanDonor and NonHumanDonor references. Cursor is providing a couple different options for doing this, can discuss what makes the most sense, but we essentially need to limit it to only checking HumanDonor and NonHumanDonor references

Jchaffer787: Can be deleted. Flattener does not do anything with these lists.

Jchaffer787: Can be deleted. Not used in flattener.

Jchaffer787 (changes requested):
Logic of the changes looks good, just went through to note what is redundant and can be deleted / adjusted.

Gatherer also returns the MatrixFileSet uuid as part of the dictionary, but it's not used in the flattener. Not sure if we still want to keep that or drop that as well?

TOOLS-285-DB2Flattener/df_utils.py
Jchaffer787: Is useful for if you want to only collapse specific columns not all, so should be kept.

TOOLS-285-DB2Flattener/DB2Gatherer.py
Jchaffer787: I am going to implement what Cursor recommended to fix this, which is a quick whitelist system to essentially limit the references we gather from samples to only be HumanDonor and NonHumanDonor.

For all other references, we will now just fill the main dataframe with whatever the reference is that's in the sample, and then afterwards go through and make report requests for the actual objects for those.

Jchaffer787:
I've added a commit removing the unused flattener and gatherer code, and making a small change to add_references_to_library in gatherer, so it's no longer adding non donor references

Jchaffer787 (approved):

PR #335, opened 2026-07-08
Tools 306 constant refactor

brian-mott:
Constants refactor to allow for parsing of DB2 JSON profiles to generate constants/configs that reflect the current schema of a given instance/endpoint of DB2. 

Before every flattener run, all DB2 JSON schema profiles are requested, parsed, and converted to the expected dictionary format for FIELD_TYPES and OBJECT_CONFIG used for gathering and flattening. For demo, prod, and sandbox, a prior version of the parsing is kept in constants.yaml to alert the user if schema changes have occurred in the generated constants; the newest parse is used regardless. 

Other changes:

- CLI arg --mode to pass DB2 instance of choice. This follows the original flattener schema to look to the local env for {mode_SERVER}, {mode_KEY}, and {mode_SECRET}. Default is db2_demo
- Early exit if mode does not start with db2_/DB2_
- Config dataclass to as container/holder for the two constants. DB2Flattener and DB2Gatherer classes now have a config instance argument and attribute to store the generated config. This allows for dynamic parsing, avoids scope issues, and avoids circular imports that occur when using pure global constants.
- Basic tests for some of the functions that help with constant generation

Tested locally with matrix file sets 9b1c06e5-ce17-4572-b243-93569b5a17e8 and 5fbe8a09-9805-4f36-9236-cb68acb145c5 on the DB2 dev/demo


Jchaffer787 (approved):
LGTM

jychien:
Amazing, looks great! Thanks, @brian-mott !

jychien (approved):

PR #336, opened 2026-07-16
TOOLS-315 SeaHub trimmed raw

gabdank:
Adds QA for collaborator lab SeaHub trimmed raw uploads (`czi-trapnell` / `czi-hamazaki`, `*-seahub-bcp`) as a new `raw_assay='seahub_sci'` mode in `qa.ipynb`. These are trimmer output rather than vendor deliveries — their own path layout, six per-well artifacts, no merged trimmer files — so none of the existing vendor-oriented checks fit, and validation is raw-only.

The SOP being enforced:

```
s3://czi-{lab}/{lastname}-{projectname}/{ExperimentID}/raw/{sublibrary}/{wafer}/
    {wafer}-{sublibrary}[_{well}]_{sublibrary type}-{UG}-{barcode}.trim.*
```

Five required artifacts per well — `.trim.cram`, `.trim.csv`, `.trim.stderr`, `.trim.stdout`, `.trim_fail.csv` — plus an optional `.trim.cram-metadata.json` sidecar.

The thing that shaped most of this: **real uploads keep arriving under different names.** REF3 dropped the `.trim` infix on six of its seven sublibraries and doubled the leading wafer token; GENE7 truncated every sublibrary folder; CHEM16 renamed the artifacts outright to `.trimmed.ucram` / `.trimmer_stats.csv` / `.failure_codes.csv`. QA's job here is to describe the gap against the SOP clearly enough that the submitter can fix it — deliberately *not* to accumulate a catalogue of every way a name can be wrong.

## What it does

Four new modules under `bcp/`, wired into a contiguous SeaHub block at the end of the notebook:

- **`qa_seahub_sop.py`** — validates path and filename structure and reports each broken rule **once per distinct fact**, at five scopes: `object`, `stem` (per well), `folder` (per sublibrary directory), `suffix` (per distinct unrecognised extension) and `upload` (per bucket/project fact). Scope is what keeps the report readable: seven truncated folders are seven rows rather than several hundred, and a wrong bucket on a 288-well upload is one row rather than 288.
- **`qa_seahub_source.py`** / **`qa_seahub_recon.py`** — reconcile the upload against the untrimmed vendor delivery listed in `untrimmed_s3_paths`, matching wells on `(wafer, UG)` and classifying `not_trimmed` / `orphan_trimmed` / `identity_mismatch` / `read_count_mismatch` / `size_not_reduced` / coverage findings.
- **`qa_seahub_rename.py`** — composes the per-rule repairs into one corrected S3 key per object, and rolls the whole upload up to **one verdict per well** (`COMPLIANT` / `RENAMEABLE` / `DATA_GAP` / `UNKNOWN`). That roll-up is the headline: the SOP table says what is wrong, the well status says how many wells are affected and which need data rather than a rename.

The rename mapping is advisory — QA never moves, deletes or rewrites anything. Every proposed key is re-validated against the SOP and withheld if the proposal is not itself clean; two objects can never be given the same destination.

Outputs: `{order}_raw_sop_violations.csv`, `{order}_seahub_well_status.csv`, `{order}_seahub_rename_mapping.csv`, `{order}_trimming_completeness.csv`, `{order}_seahub_source_coverage.csv`.

Also in scope: per-format trim-failure percentages (the shared single-denominator parser produced >100% on these files), a reads-per-wafer row for every discovered wafer, and `parse_raw_filename` returning the folder sublibrary rather than the ExperimentID so per-sublibrary aggregation stops collapsing an experiment into one row.

## Design decisions worth a reviewer's attention

**Wells are matched on `(wafer, UG)`.** Measured on a real delivery, each wafer carries exactly 48 CRAMs with 48 distinct UGs. That pair survives every naming defect seen in review — duplicated wafer tokens, missing type tokens, vendor-vs-lab group formatting — whereas the filename does not. Barcode and sublibrary type are carried as verification fields and reported on mismatch rather than being allowed to break the match.

**Completeness asks which artifact *kinds* arrived, not what they were called.** Naming is a separate axis, reported once per stem as `missing_trim_infix`. Judging by suffix family instead let a single optional sidecar carrying the other family's name decide the whole requirement set — and the two completeness paths reacted differently, so a well with five correct `.trim.*` files plus a stray bare sidecar was complete to the roll-up and missing five files to the inventory. They now agree by construction. A genuinely absent artifact still reports, under its SOP name whichever spelling the upload used.

**The vendor index is scoped to the ExperimentID being QA'd.** A listed prefix is order-level and one order holds several experiments, so without the filter the other experiments' wells are indexed too and each reports as `not_trimmed` plus an `UNKNOWN` well. Two segment shapes count as a match — the ExperimentID alone (`REF3`) and the older `{ExperimentID}_{sublibrary}` (`REF5_P01`) — because a bare equality test would orphan every well of an old-shape delivery. The filter reads `ctx.order`, not the notebook's `order`, which is `ctx.output_label` and therefore whatever `run_label` was set to; filtering on a run label matches no vendor well and empties the index.

**A wholly misnamed upload fails loudly.** `unexpected_suffix` collapses to one row per distinct extension naming the count, an example and the SOP artifacts expected; if nothing at all is recognised, `no_recognized_artifacts` fires once and the notebook prints a FAIL banner before any per-well table, since no well can be identified and every table below would otherwise be empty and read as clean. A bare suffix is only accepted when the stem before it parses — `.csv` is generic enough that `<well>.trimmer_stats.csv` otherwise becomes its own well, which turned CHEM16's 480 wells into 960.

**No alias table.** `.trimmed.ucram` is not mapped to `.trim.cram`. Chasing each new way an upload can be wrong is unbounded; describing the gap against the SOP is not.

## Verified against three real listings

Run offline from S3 listings of `czi-trapnell`. Every number below can be re-derived by anyone
holding the same listings — no S3 access, no credentials:

```bash
python -m tests.seahub_offline_report LISTING.tsv --mode both
```

`tests/seahub_offline_report.py` drives the same production code the notebook does, with
`MockS3Client` standing in for S3. Nothing identifies the upload on the command line: the
ExperimentID is read out of the listing, which also exercises the manifest-mode recovery. Expected
output for all three uploads is in `bcp/README.md`, and `--json` gives a diffable snapshot.

| upload | objects | SOP rows | wells | completeness |
| --- | --- | --- | --- | --- |
| REF3 | 2037 | 892 | 336 (48 `RENAMEABLE`, 288 `UNKNOWN`) | 336 stems, none missing an artifact |
| GENE7 | 5185 | 10 | 864, all `RENAMEABLE` | 864 stems, none missing an artifact |
| CHEM16 | 1450 | 14 | 0 | n/a — no well identifiable |

REF3's 892 rows are 288 each of `duplicated_wafer_token`, `invalid_sublibrary_type` and `missing_trim_infix`, 21 browser-download leftovers classified as junk, and 7 truncated sublibrary folders. Its 288 `UNKNOWN` wells are an *input* gap, not a data one: those filenames dropped the sublibrary type and only the vendor delivery can supply it, so they resolve once `untrimmed_s3_paths` is set.

GENE7 is one folder rename from clean — 9 truncated sublibrary folders (`P02` → `GENE7_P02`) and one 5.3 GB extensionless object at `GENE7/raw/P10/436516`, flagged `bad_path_depth`.

Every row above reproduces identically from an S3 listing and from a manifest of the same keys — same objects, SOP rows and rules, wells and verdicts, rename statuses. That parity is asserted in the suite, not just observed: the two modes are the same question asked twice.

CHEM16 produces 14 rows rather than the 2410 it did before the suffix work: 10 junk, 3 naming rows (one per spelling, each with its count and an example) and the one headline failure.

### Reconciliation against the vendor delivery

Run against the two Novogene orders REF3 was trimmed from — `NVUS2024101701-02` (48 wells) and `NVUS2024101701-11` (288 wells), both order-level prefixes:

- **336 vendor CRAMs indexed, 336 trimmed wells, 336 matched.** No `not_trimmed`, no `orphan_trimmed`, no `duplicate_source_well`. Trimming is complete for this upload.
- **Supplying the vendor index resolves every `UNKNOWN` well.** The per-well status goes from 48 `RENAMEABLE` / 288 `UNKNOWN` to **336 `RENAMEABLE`, 0 `UNKNOWN`, 0 `DATA_GAP`** — confirming those wells were an input gap, not a data one.
- **The size check ran on all 336 wells** with real sizes: every trimmed CRAM is 48–70% of its vendor source (median 64%, 23.6 GB against 37.7 GB), so nothing flagged. That is the check that would catch untrimmed data landing in the trimmed bucket.

One caveat on this run: read counts live inside the vendor sidecar JSONs, which a listing does not carry, so all 336 wells report `metadata_unavailable`. A live notebook run downloads them via `load_source_read_counts` and the read-count check does execute.

The one missing CRAM the original review found (wafer 438514, well A3) is present in this listing with all six artifacts, which is why completeness now reports no gap.

### The ExperimentID filter, measured on an order holding two experiments

Order `NVUS2024101701-11` holds `REF3` (288 wells) beside `GENE7_reupload` (864). Reconciling the REF3 upload against that order-level prefix:

| | vendor wells indexed | `not_trimmed` | `UNKNOWN` wells |
| --- | --- | --- | --- |
| without the filter | 1152 | **864** | **864** |
| with the filter | 288 | 0 | 0 |

So the 864 false completeness failures and 864 phantom wells are what the filter removes, on real data. The 48 `orphan_trimmed` that remain in both columns are correct: those wells came from order `NVUS2024101701-02`, which this prefix list does not include — the documented "a forgotten order looks like a completeness failure" case, which the coverage table exists to distinguish.

That second folder is named `GENE7_reupload`, not `{ExperimentID}_{sublibrary}`. The two-shape match already accepted it, since the rule is any `{ExperimentID}_...` folder, but it had been justified and tested only on the older sublibrary shape — hence a regression test for the re-upload shape, including that `GENE7_reupload` must not match `REF3`.

**GENE7 reconciles clean against the same order:** 864 vendor wells, 864 trimmed, 864 matched, no `not_trimmed`, no `orphan_trimmed`, all 864 `RENAMEABLE`. Its trimming is complete; the truncated folders are its only defect.

## Tests

1339 passing. Fixtures and cases are drawn from the real listings, with lab and order identifiers sanitized. Both reviewer-supplied known-good filenames validate clean, including the hyphenated-ExperimentID form with no well token, and a compliant upload proposes no renames.

`SEAHUB_SOP_RULES` claims a typo in a new rule shows up as a test failure. Asserting over the *emitted* types does not deliver that — the fixtures do not exercise every rule, and a misspelled `wafer_mismatch` passed when tried. The rule and scope vocabularies are checked by reading the `SopViolation` literals out of the module with `ast`, in both directions, so a typo and an unreachable rule each fail.

## Review follow-up

Twenty-three commits over five review rounds, one per item. Each was verified against the three real
listings, carries a test that fails without it — checked by reverting the source and re-running —
and is explained in its own commit message.

**Round 1 — two blocking, four should-fix, five nits, three decisions.** Manifest mode hard-coded an
empty ExperimentID, which reached the cross-experiment check as the *expected* value and made it
true for every object: 2037 false errors on REF3, 5184 on GENE7, 1450 on CHEM16. It also silently
disabled the vendor index's experiment filter, reproducing the regression `ae89f00` had just fixed.
Separately the S3 walk kept only `CommonPrefixes` and discarded `Contents`, so any object above
wafer depth was invisible — `bad_path_depth` could fire only for keys too *deep*, and the two modes
disagreed about the bucket. Both were green because the tests hand-built a context the resolver
cannot produce and hand-wrote listings that could not express a stray object. Also: the two headline
CSVs could contradict each other (one predicate decides moveability now), `METADATA MISSING` was one
warning per CRAM, and dead vocabulary went, with an `ast` guard against the next.

**Decisions taken with the author.** A vendor well with nothing uploaded is `DATA_GAP`, not
`UNKNOWN` — the largest gap there is, and the only one the notebook left out of `errors.txt`.
Upload-scope rules no longer veto the rename plan: `bad_bucket` and `lab_project_mismatch` are
invariant under a move, so one wrong bucket withheld every proposal in the upload; the plan is
produced with a banner saying the destinations sit somewhere the SOP rejects. `repeated_token` is
scoped to the sublibrary portion of a matched stem, so a sublibrary legitimately named
`REF3_GEX_P01` under type `GEX_hash_oligo` no longer flips a whole sublibrary to `UNKNOWN`.

**Round 2 — no blocking.** `reconcile_trimming` tallied matches into its caller's `SourceCoverage`
objects, so re-running the recon cell without re-running the indexing cell above it doubled
`matched` and floored `unmatched` at zero — a source with genuinely unmatched wells reporting as
fully accounted for. One unreadable vendor sidecar took the whole cell down; it degrades to
`metadata_unavailable` now. And the per-well trim failure CSVs were fetched serially — 336 sequential
round-trips on REF3, 864 on GENE7, while the sidecars beside them used a 16-worker pool. They share
that pool now: **10.6s to 1.2s** at 20 ms per object, with the parsed blocks applied single-threaded
in listing order so the result is byte-identical rather than merely equivalent.

**Round 3 — no blocking.** A gap row carried no `sublibrary` or `well`, making the highest-severity
verdict the least locatable row in the headline CSV. `lab_project_mismatch` compared the first
hyphen-separated token, so a lab whose name contains a hyphen failed against its own project.
`WELL_COLUMNS` was dead, and the guard meant to catch that was scoped too narrowly to see it.

**Round 4.** The remaining nits, and `qa_checks.py` keyed its completeness `beginnings` on the stem
alone — so a well uploaded under two folders read as one complete well plus five unexpected files.

**Round 5 — three false all-clears.** A 0-byte CRAM reported `COMPLIANT` (an interrupted copy leaves
exactly that shape); one malformed per-well file aborted the whole gather, losing the 863 that were
fine, in both the trim-CSV and the metadata-sidecar pool; and an upload in the wrong bucket or
project reported every well `COMPLIANT` with nothing in `errors.txt`, because the warning banner was
gated on there being something to rename and an otherwise-clean upload has nothing to rename.

**On verification.** From round 3 on, the diff went through adversarial review before being pushed,
and it changed the outcome every time: it caught a first-cut fix that put a wrong value in the
headline CSV where a blank had been, two dead-code guards that were exempting the very constants
they were written for (both matched other files as raw text, and both name those constants in their
own docstrings), four tests that passed against mutations they should have caught, and — on the last
round — a fix that would have fabricated a whole missing well out of a stray object and dropped that
object from the extras list.

Numbers on REF3, GENE7 and CHEM16 are unchanged throughout except where a fix is the point: GENE7
gains the `bad_path_depth` row it should always have had.

## Out of scope / follow-ups

- Upload checksum verification stays manual; the read-count reconciliation is the automated substitute. S3 validates integrity on upload.
- That 5.3 GB extensionless GENE7 object needs identifying by a human.


bcp/qa_constants.py
```diff
@@ -221,6 +229,13 @@
...
     ],
     "scale": [],
+    "seahub_sci": [
+        ".trim.cram",
+        ".trim.csv",
+        ".trim.stderr",
+        ".trim.stdout",
+        ".trim_fail.csv",
+    ],
 }
```
jychien, line 239: @gabdank I ran the qa notebook with s3_path = `s3://czi-trapnell/trapnell-seahub-bcp/REF3`:
- *_raw_extra.txt was correctly listing out the extra files. Ex: `trapnell-seahub-bcp/REF3/raw/P04_1/437120/437120-437120-REF3_P04_1_A1-Z0001-CAGCTCGAATGCGAT.cram`. 
- However, I did not see any of the required files as missing. The actual file name should be `trapnell-seahub-bcp/REF3/raw/P04_1/437120/437120-437120-REF3_P04_1_A1-Z0001-CAGCTCGAATGCGAT.trim.cram`, as is in this list.

jychien (changes requested):

jychien:
@gabdank Here are some issues/question I found:
- For `trapnell-seahub-bcp/REF3/raw/P04_1/437120/437120-437120-REF3_P04_1_A1-Z0001-CAGCTCGAATGCGAT.cram`, the wafer ID is listed twice in the file name.
- It would be great to additionally be able to check that all of the original untrimmed files in `czi-novogene` are all accounted for. And comparing read counts would be an additional check.
- Is the reads per wafer table in the qa notebook supposed to be reporting all the wafers? Currently, I am only seeing wafer 430479.
- For the Hamazki upload, Brent has uploaded checksums, and I will be double checking those as well to make sure that the full file was uploaded.

I am not sure if we need to add seahub trimmed raw upload to this same production qa notebook, if that helps simplify things.

gabdank:
@jychien this is fixed — thanks, it was the right catch and it turned out to be the more interesting half of a real inconsistency.

You were seeing one file counted twice over: the bare `.cram` was reported as **extra**, while the `.trim.cram` it stands in for was *not* reported as **missing**. Those two answers cannot both be right. The cause was that completeness judged a well by suffix *family*, so a single sidecar carrying the other family's name could decide the whole requirement set — and the two completeness paths disagreed about which five files were required.

Completeness now asks only whether each of the five artifact **kinds** arrived, under either spelling. Naming is a separate axis, reported once per stem as `missing_trim_infix`. On your exact object, from the same REF3 listing:

```
found  P04_1/437120/437120-437120-REF3_P04_1_A1-Z0001-CAGCTCGAATGCGAT.cram
found  ....csv   found  ....stderr   found  ....stdout   found  ...._fail.csv
```

- `raw_extra` = 21, and none of them is a required artifact — those 21 are the browser-download leftovers (`login.html`, `ug-icon.png`, …).
- `raw_lost` = 0, `all_good` = 336 wells.
- The naming problem is still reported, in the SOP table: `missing_trim_infix` × 288, alongside `duplicated_wafer_token` × 288 and `invalid_sublibrary_type` × 288.

So the file is counted as present, and the fact that it is misnamed is its own row rather than a phantom "extra".

Three further review rounds landed since yours — 22 commits in total, each with a test that fails without it. The PR description has the summary. You can re-derive any number in it yourself, without S3 access, from a saved listing:

```bash
python -m tests.seahub_offline_report LISTING.tsv --mode both
```

Re-requesting your review. CI is green (`bcp-pytest`, `lint`, `review`), 1316 tests passing.

gabdank:
@jychien answering your two open questions, and reporting three fixes that came out of reviewing the final state.

**"Do we need seahub trimmed raw in this same production notebook?"** — yes, it belongs here. It is just another mode of raw data submission, and it should not be a separate tool. On the bloat concern: the SeaHub block is 10 of 27 cells but **52% of the notebook's lines**. What keeps that tolerable is containment — the cells are contiguous at the end, every code cell is gated on `seahub_active`, and `tests/test_qa_notebook_layout.py` pins both properties so an assay-specific cell cannot drift into the shared pipeline again. A non-`seahub_sci` run prints "skipped" ten times and does nothing else.

**Hamazaki checksums** — not needed. S3 computes CRC64NVME on upload, so transfer integrity is AWS's responsibility rather than QA's. The PR's read-count reconciliation covers the question we actually care about: did trimming consume the whole delivered file.

**Three fixes since your review**, all of the same kind — the headline output could read clean on a broken upload:

- **A 0-byte CRAM reported `COMPLIANT`.** Completeness only asked whether the artifact arrived, and the size check skipped a zero because it could not tell "empty" from "size not collected". An interrupted `aws s3 cp` leaves exactly that. It is a `DATA_GAP` now, in the headline CSV and in `errors.txt`. Only the CRAM — an empty `.stdout` is normal.
- **One malformed file aborted the whole run.** There is one trim-failure CSV and one metadata sidecar per well, both parsed unguarded in a thread pool, so one zero-byte or half-written object out of 864 lost the other 863. Both degrade now, with the failures collapsed into a single warning.
- **An upload in the wrong bucket or project reported every well `COMPLIANT`** and wrote nothing to `errors.txt`, because the warning banner was gated on there being something to rename — and an otherwise-clean upload has nothing to rename. That gate is gone.

1339 tests passing. The numbers in the PR description are unchanged, and you can re-derive any of them from a listing without S3 access:

```bash
python -m tests.seahub_offline_report LISTING.tsv --mode both
```

Known and deliberately left for a follow-up, so they are not a surprise: `duplicate_trimmed_well` reports once per artifact rather than once per well (6x row inflation, no lost information); a well whose artifacts are *all* misspelled drops out of the per-well table rather than showing as a gap; and `errors.txt` is append-only, so a stale line can survive a re-run.

bcp/qa.ipynb
```diff
@@ -1,671 +1,1442 @@
...
+    "`[ACGT]+`, so a malformed UG or barcode cannot match either pattern and is\n",
+    "reported as `unparseable_stem`, whose detail names the expected shape.\n",
+    "\n",
+    "A folder/filename disagreement is blamed on the **folder** when the filename says\n",
+    "`{ExperimentID}_{folder}`, because the vendor delivery is the source of truth for\n",
+    "the sublibrary name. `sublibrary_mismatch` is reserved for names the folder\n",
+    "cannot explain either way.\n",
+    "\n",
+    "Violations are non-fatal. Full detail goes to\n",
+    "`{output_dir}/{order}_raw_sop_violations.csv`."
```
jychien: This description is confusing to read. I think the table summary might be good enough. I like how full details are in a separate doc if people have questions.

bcp/qa.ipynb
```diff
@@ -1,671 +1,1442 @@
...
+    "\n",
+    "    trimmed_index = index_trimmed_upload(\n",
+    "        all_raw_files,\n",
+    "        sizes=raw_file_sizes,\n",
+    "        findings=trimmed_dupes,\n",
+    "    )\n",
+    "    print(f\"  {len(trimmed_index)} trimmed well(s) indexed by (wafer, UG)\")\n",
+    "    if trimmed_dupes:\n",
+    "        print(f\"  WARNING: {len(trimmed_dupes)} well(s) appear under more than one name\")"
+   ],
```
jychien, line 919: I ran with:
`untrimmed_s3_paths = ["s3://czi-novogene/hamazaki-seahub-bcp/NVUS2024101701-09/",
                     "s3://czi-novogene/hamazaki-seahub-bcp/NVUS2024101701-34/",
                     "s3://czi-novogene/hamazaki-seahub-bcp/NVUS2024101701-37/",
                     "s3://czi-novogene/hamazaki-seahub-bcp/NVUS2024101701-59/"]`

This cell resulted in:

> NVUS2024101701-09          0 indexed  (0 CRAM key(s) seen)
    WARNING: s3://czi-novogene/hamazaki-seahub-bcp/NVUS2024101701-09/ yielded no vendor CRAMs

However, there are corresponding wafers in `s3://czi-novogene/hamazaki-seahub-bcp/NVUS2024101701-09/R102/raw/`:
- 435934
- 436053
These two wafers are listed as missing untrimmed source. I am guessing this probably is because the "experimentID" is different. When I did a quick qa on this data, what I did was just go through all of hamazaki's orders, and look for the wafer#. This is because the "experimentID" is unreliable. However, I do realize this does not catch wafers that are part of the experiment but missed in their trimming.

PR #338, opened 2026-07-23
Tools 313 finalizing main df

Jchaffer787:
- Adjusted the reference handling to gather all reference objects assigned by constants, and still perform batch gathering by object type. 

- Changed the handling in Flattener.py to not filter GEX Libraries for the MAIN dataframe. To filter GEX files, follows previous implementation of filtering.

- Added filtering of GEX libraries for the GEO dataframe.

- Addressed last donor overwriting bug. Values are now made into lists if there’s multiple samples in the donor.

- Added a check for ‘author_metadata’ field during creation of the MAIN dataframe, and adjusted how the column header and value is set accordingly.

- Adjusted creation of OBJECT_CONFIG by filtering out '_creation_timestamp' and '_submitted_by' fields, so they don't end up in the main dataframe.

### DIFF
--- a/TOOLS-285-DB2Flattener/DB2Flattener.py
+++ b/TOOLS-285-DB2Flattener/DB2Flattener.py
@@ -210,7 +210,7 @@ class DB2Flattener:
         return main_df, new_sample_df
 
     def _row_is_gex(self, row) -> bool:
-        """ Filter MAIN df to only GEX libraries"""
+        """ Filter df to only GEX libraries"""
         droplet_ft = row.get('droplet_based_libraries_feature_types')
         plate_ft = row.get('plate_based_libraries_feature_types')
 
--- a/TOOLS-285-DB2Flattener/DB2Flattener.py
+++ b/TOOLS-285-DB2Flattener/DB2Flattener.py
@@ -4,7 +4,13 @@ import pandas as pd
 from datetime import datetime
 from DB2Gatherer import DB2Gatherer
 from constants import Configs, PROP_MAP_GEO
-from df_utils import split_controlled_term_columns, collapse_dataframe
+from DB2_utils import (
+    split_controlled_term_columns, 
+    collapse_dataframe,
+    get_config_obj_type,
+    get_url_prefix_from_id,
+    extract_references_from_field
+)
 from generate_constants import load_and_return_constant_dicts
 import DB2lattice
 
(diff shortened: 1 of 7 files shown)


Jchaffer787:
Do we want to implement these lint ruff test findings? Looks like not all of them are regarding the flattener code.

TOOLS-285-DB2Flattener/DB2Flattener.py
```diff
@@ -188,63 +144,64 @@ def create_dataframe(self, complete_data):
...
-
-                        for d in sample_donors:
-                            donor_type = get_config_obj_type(d, self.configs)
-                            for field in self.configs.OBJECT_CONFIG[donor_type].get('fields', []):
-                                field_name = f'{donor_type}_{field}'
-                                value = d.get(field)
+                            # Special handling for author_metadata dictionary
+                            if field == 'author_metadata' and isinstance(value, dict):
+                                for key, val in value.items():
+                                    field_name = f"{sample_type}_{'_'.join(key.split(' '))}"
```
jychien: @Jchaffer787 I was thinking more about this, and realized that if we named them as such (example: `tissues_mouse_litter_batch`), we will lose the fact that this was an author column. If we retained the string "author_column", then, for example, the GEO csv generator would know which column names are author columns to retain for the GEO csv. So, I was thinking to retail "author_column" in the field_name, and end up with `tissues_author_metadata_mouse_litter_batch`. Thoughts?

TOOLS-285-DB2Flattener/DB2Flattener.py
```diff
@@ -259,20 +216,113 @@ def create_dataframe(self, complete_data):
...
+                    if value is None:
+                        continue
+                    if is_controlled_term:
+                        value = self._resolve_controlled_term(
+                            value, resolved_controlled_terms
+                        )
+                    if value not in (None, ''):
+                        values.append(value)
+
+                sample_metadata[sample_alias][col] = (
```
jychien: MatrixFileSet `9b1c06e5-ce17-4572-b243-93569b5a17e8` unintentionally has author_metadata of mouse_litter_batch at both donor and sample objects. So, the main_df still has column `non_human_donors_author_metadata`, with value `{'mouse_litter_batch': 'JSS1'}`. This portion of the code will need to check if col == `author_metadata`, and then do the same thing as for sample, where column name and value is the keys and values found in the author_metadata dictionary.

jychien (changes requested):

TOOLS-285-DB2Flattener/DB2Flattener.py
Jchaffer787: Agreed. I will add that.

Jchaffer787:
@jychien I've added the changes!

jychien:
@Jchaffer787 Looks good to me, thank you! I do see the lint failing check. Is there some formatting issue?

jychien (approved):

Jchaffer787:
@jychien I'm looking through it, and it doesn't look relevant to this, it's all in the bcp folder

PR #339, opened 2026-07-29
Tools 317 add sequence file and sequence file set metadata to main

Jchaffer787:
In Gatherer.py, instead of only grabbing the library_uuids when traversing the raw_matrix_file -> sequence_file -> sequence_file_set -> library path, it also grabs the sequence_file and sequence_file_set information. Then, instead of appending the raw file data to the library data through the library_uuuid, it appends a shallow copy of the raw file data, which also includes the sequence_file and sequence_file_set information.

In Flattener.py, when creating rows for the dataframe, instead of having the same raw file object metadata for each library within the raw file, it assigns the specific shallow copy version of that raw file that has the sequence_file and sequence_file_set information specific to that library attached.

The result is that the MAIN dataframe still is by {raw_file, library}, but the sequence_file and sequence_file_set information is now included, and also specific to each library.


gabdank:
@Jchaffer787 rebase onto main and the tests will pass

PR #342, opened 2026-07-31
TOOLS-321 Add guidesig protospacer set signature tool

gabdank:
## Summary

Adds `guidesig`, a stdlib-only library and CLI that computes a deterministic `gsig1:set:n=…:…` signature over the set of protospacer sequences in a guide-template file, so two libraries can be checked for identity without eyeballing them. It is a string-identity check on one column, not a biological equivalence check: length, strand, and orientation are deliberately not normalized, because a false match is the failure mode this tool exists to prevent.

Both TSV and CSV input are supported, and the format must be stated explicitly — `--format tsv|csv` on the CLI, `file_format=` in the library. Nothing is inferred from the file extension or sniffed from the contents, since sniffing misclassifies the sparse single-column rows these templates contain and a silent misparse is exactly what must not happen.

**Changes:**
- `bcp/guidesig/` — `core.py` (parsing, validation, hashing), `cli.py` (signature, `--compare`, `--column`, `--format`), and `__main__.py` for `python -m guidesig`
- Signatures are format-independent: the same sequences as TSV and as CSV produce the same digest, pinned by a CSV conversion of the layout A fixture that asserts the existing TSV digest
- Rejects ragged rows, non-ACGT values, a missing signature column, invalid UTF-8, and files with no usable protospacers; ignores blank rows, `#` comment rows, and empty protospacer cells
- Rejection messages report physical file line numbers, so a quoted cell containing an embedded newline does not shift the line the user is pointed at
- Golden fixtures plus 67 unit, regression, and CLI tests across `test_guidesig.py`, `test_guidesig_csv.py`, and `test_guidesig_cli.py`
- Documents the guarantees and the deliberate non-guarantees in `bcp/README.md`

**Breaking changes:** none for existing tools — `guidesig` is new in this PR. Within the branch, requiring `--format` means `python -m guidesig lib.tsv` exits 2 and `signature(path)` raises `TypeError`.

## Test plan

- [x] `pytest tests/test_guidesig.py tests/test_guidesig_csv.py tests/test_guidesig_cli.py` — 67 passed
- [x] Full `bcp` suite — 758 passed, 13 skipped
- [x] `ruff format --check` and `ruff check` clean
- [x] Same library in both formats reports one digest: `python -m guidesig --format csv tests/fixtures/lib_two_layout_a.csv` and `--format tsv …a.tsv` both give `gsig1:set:n=60:0ae925412d48830d0e5bee973c24ea78`
- [x] `--compare` on `lib_two_layout_a.tsv` vs `lib_two_layout_b.tsv` reports match
- [x] `--compare` on `lib_one.tsv` vs `lib_one_prepended_g.tsv` reports mismatch and exits non-zero


brian-mott:
I'm going through logic, reasoning, and tests on other metadata guide files I have used, so far so good on those fronts.

For the CXG 7.1 validation work, I created CSV files instead of TSV files for the fixtures. I don't know if I picked a specific reason for this. Should we have some flexibility to check either TSV or CSV files? Is TSV more standard/required for the curated matrix ingest or is there some other reason for selecting TSV over CSV?

gabdank:
No good reason to pick TSV over CSV, we can expand support to both.

brian-mott:
What do we envision the use of the --compare flag/feature? Like if we are submitting to DB2 and think there might be an equal or almost equal file, we can use this for an initial check of similarity? The check itself makes sense but currently won't work if we are comparing between a CSV and TSV file. Not sure if we want/need to do this or if that will make things more complicated to sniff/detect file formats instead of just providing the --format arg

gabdank:
Re: the `--compare` question — what it's for, and the CSV/TSV restriction.

### What `--compare` is for

Roughly the split you describe, with one wording correction. The two modes answer different questions:

- **`signature`** is the identity you *store*. Compute it at submission time, keep it on the fileset record or in the ticket, and "does DB2 already have this guide set?" becomes an O(1) lookup instead of a pairwise guess. That is the durable answer for the submission workflow, and it's follow-up schema work rather than anything in this PR.
- **`--compare`** is the diagnostic you reach for *after* two signatures disagree and someone has to explain why. `|A|`, `|B|`, `|A∩B|`, `|A−B|`, `|B−A|` plus example sequences distinguish "someone added 10 guides" from "every guide lost its leading G" — two situations that a bare signature mismatch reports identically.

So: signatures answer **whether**, `--compare` answers **how**.

One thing worth being explicit about, since "almost equal" could be read into the output: this is not fuzzy or approximate matching. `--compare` reports exact set difference. A library where every protospacer is trimmed by one base has `|A∩B| = 0` and reads as maximally different. That's the intended answer, not a limitation — a trimmed-base or strand-flipped file producing a *near*-match would be exactly the false reassurance the tool exists to prevent.

### The CSV-vs-TSV restriction

Correct that `--compare` applies one `--format` to both files. The gap is narrower than it looks, though, because signatures are already format-independent: `signature()` hashes only the sorted protospacer set, never the delimiter, column order, or layout. So cross-format comparison works today in two invocations:

```bash
python -m guidesig --format csv lib_a.csv
python -m guidesig --format tsv lib_b.tsv
```

Identical signature portions → identical protospacer sets. As a one-liner that gives a straight yes/no (the `cut -f1` drops the path column, which always differs):

```bash
diff <(python -m guidesig --format csv lib_a.csv | cut -f1) \
     <(python -m guidesig --format tsv lib_b.tsv | cut -f1) && echo "same set"
```

This is already pinned by `test_csv_fixture_matches_tsv_golden_digest`: `lib_two_layout_a.csv` and `lib_two_layout_a.tsv` assert the same digest, which is what proves the two readers agree.

What the workaround loses is only the diagnostic detail and the non-zero exit code on mismatch — not the yes/no answer.

### On sniffing: no

Content sniffing is the one option I'd rule out. It's what `abf39f3` deliberately rejected: sniffing misclassifies the sparse single-column rows these templates contain, and a silent misparse is the failure mode this tool exists to prevent. Adding detection only for `--compare` would reintroduce that risk through a side door and leave the CLI explicit in one mode and magic in the other.

If we do want cross-format `--compare`, the explicit version is cheap — an optional `--format-b` defaulting to `--format`:

```bash
python -m guidesig --format csv --compare a.csv b.tsv --format-b tsv
```

~15 lines in `cli.py` plus README, and the test fixtures already exist (the pinned `lib_two_layout_a` CSV/TSV pair). No ordering trick, and single-format behavior is untouched.

My read: for the duplicate-check use case the two-signature route is genuinely sufficient, since that question is yes/no. Cross-format `--compare` only matters once a cross-format mismatch actually shows up and needs explaining — and converting one file then re-running `--compare` covers that too. So `--format-b` is a convenience, not a capability gap. Happy to add it in this PR or document the workaround and leave it; no strong preference.

gabdank:
Closing the `--compare` thread: **`--compare` stays single-format, and we're not developing it further for now.** It's a nice-to-have diagnostic, not the point of the tool — the feature is the signature. So no per-file format argument, and no content sniffing.

What that leaves is the mixed-format case, where a clear message is enough. Reading a file under the wrong format previously reported only:

```
error: lib_b.tsv: protospacer column 'guide_protospacer' absent from header
```

Accurate, but silent about the cause. It now names the format (`bb6aae4`):

```
error: lib_b.tsv: protospacer column 'guide_protospacer' absent from header; the whole
header parsed as one csv field, so this file looks like tsv rather than csv (check the
format argument)
```

The hint fires when the header collapsed to a single field **and** still holds another format's delimiter — the fingerprint of a format that doesn't match the file. Both conditions are required, so a genuinely wrong `--column` on a well-formed file isn't blamed on the format, and neither is a legitimately single-column file; both are pinned by tests. A hint firing on the wrong cause would be worse than no hint.

It lives in the parser rather than in `--compare`, so it also covers the more common case of running `--format csv` over a TSV by hand. Wiring a "both files must share a format" note into the CLI's error path would have printed on every compare failure — unreadable file, non-ACGT value — where it's just noise.

For the record, if cross-format comparison is ever actually needed: signatures are already format-independent, so signature each file under its own `--format` and read the two lines. No new flag required.

### Also pushed since the review

- `ad59689` — rows carrying values in fields **beyond** the header are now rejected (the open question in the review). A surplus *empty* field is a trailing delimiter, which exports produce constantly, so that stays tolerated; a surplus *value* means a truncated header or shifted export, and the protospacer read from such a row can't be trusted. Sits with the existing short-row check, so ragged classification still precedes the empty-value skip and a sequence shifted out of the protospacer column is rejected rather than skipped.
- `f79de4f` — the three nits. README section on why the digest is 128 bits (accidental divergence, not adversarial collision — explicitly not tamper evidence, since anyone who can edit a library can recompute its signature; `gsig2` named as the escape hatch if that ever changes). Comment-marker scoping documented in both the README and the parser, including why the first-column coupling is safe: a template with `property` moved off the front loses comment filtering but then fails loudly at a named line rather than matching falsely. Pinned by a `#` in a later column staying data, and by `#` added to the non-ACGT parametrization.
- One correction on the `_delimiter_for` note: `file_format=None` takes the `KeyError` path, not `TypeError` — `None` is hashable. Only an unhashable value reaches the `TypeError` arm. Both are now covered by tests, and a comment records that neither is reachable from the CLI.

Test count is 78, up from 67 at review time.

brian-mott (approved):
Ok I think that's fine for now to not have cross-format compare. Seems like Claude can whip up an implementation easily enough if through use we find that will be helpful. Otherwise, this looks good; signatures that include the sorted protospacer sequences and number of guides should be a solid start for comparison and the --compare flag will likely be helpful in the first step to investigate how similar two files are to one another.

PR #343, opened 2026-07-31
Tools 317 add sequence file and sequence file set metadata to main

Jchaffer787:
Adjusted _map_raw_matrix_files_to_libraries in gatherer.py to also store the sequence_file and sequence_file_set used to get to that library in the mapping. This way, when going back through the mapping to add raw matrix file objects to the library data, it adds the specific sequence_file and sequence_file_sets along with the raw file. 

Changing how create_dataframe adds rows to the dataframe. Now as it iterates through the raw files, and libraries within those raw files, it will grab the specific raw file copy (which contains the library specific sequence_file and sequence_file_set information), and add that for the library, instead of just the general raw file object.


Jchaffer787:
@jychien I also rebased this to main to fix the lint test issues.

PR #346, opened 2026-08-05
add biohub csv flattener, including utils function to combine time bo…

jychien:
…und fields and contants for mapping


jychien:
@Jchaffer787 Okay, finally ready for review. 

I ended up adding disease for MatrixFileSet `9b1c06e5-ce17-4572-b243-93569b5a17e8` for testing. It looks like `_get_diseases_from_samples()` and only store term ID. Is this being used by another flattening function?

gabdank:
@jychien  if you rebase onto main I think the PR reviews from Claude should be "fixed" (removed except for the last one)

TOOLS-285-DB2Flattener/DB2Flattener.py
```diff
@@ -279,6 +300,72 @@ def create_geo_dataframe(self, main_df) -> pd.DataFrame:
...
+        biohub_df = combine_bound_columns(
+            biohub_df,
+            lower_col="tissues_lower_bound_age",
+            upper_col="tissues_upper_bound_age",
+            units_col="tissues_age_units",
+            out_col="age",
+        )
+
+        # Update values to match schema
+        biohub_df["tissue_type"] = biohub_df["tissue_type"].apply(lambda x: TISSUE_TYPE_MAP.get(x[0], x))
```
Jchaffer787, line 359: @jychien We might want to add a datatype check here for 'tissue_type' value before checking it against the tissue_type_map. Here we are assuming that the values in the 'tissue_type' cells are all tuples, but I don't think we can safely make that assumption here. If the value in "tissue_type" is a string, this mapping will not work correctly.

jychien: In line 318, the script turns all list columns in biohub_df into tuples. This is to allow for dropping of duplicate rows, since lists are not hashable and I was unable to drop duplicates with lists.
`tissue_type` is converted from `tissue_@type`, and `@type` is a non-submittable property from all objects, so it should at least be a list of 1 length. But please let me know if I've missed something.

jychien:
@Jchaffer787 I've added `3d79e430-77a6-42d6-b7b6-048e1f005935` to prod. This is ready for final review. Thanks!

TOOLS-285-DB2Flattener/DB2Flattener.py
jychien: @Jchaffer787 I've just asked Idan, and he says that for DB2, tissue_@type will always be an array, so I think we'll be okay.

